Issue #649, opened 2014-12-19
App name editing UI missing dashed border in Editor


flukeout:
![image](https://cloud.githubusercontent.com/assets/25212/5511513/6b32ab3e-8783-11e4-94bf-bbfdbe1c549c.png)

It's supposed to have a blue dashed border underneath it - but somehow it got lost.

k88hudson:
This was resolved

PR #660, opened 2015-01-06
adding underline

gvn:


gvn:
Resolves #649

gvn:
R? @k88hudson

PR #682, opened 2015-01-09
Fixes #670—Delete button on edit page should match block editor delete button

### LINKED ISSUE #670: Delete button on edit page should match block editor delete button
i.e. on `/make/:id` the large red delete button should look like this instead:

image

gesa:


k88hudson:
Thanks!

PR #692, opened 2015-01-12
Cordova integration

k88hudson:


lib/router.js
```diff
@@ -59,6 +61,11 @@ module.exports = function (app) {
         app.params = e.params;
     });
 
+    page('/', '/sign-in');
+
+    // Cordova
+    page('/:platform/www/index.html', '/sign-in');
```
k88hudson: this is pretty gross hack but seems to be necessary on first load

static/index.html
```diff
@@ -2,11 +2,12 @@
 <html<% if (APPCACHE) { %> manifest="manifest.appcache"<% } %>>
 
     <head>
+        <base href="/android_asset/www/">
```
k88hudson: I'm not really sure how else do to this.

There are a lot of paths that need to be relative (fonts, images, etc.). 

I'm sure this would need to be re-written for ffos and ios.

k88hudson: Also, for pure web, the base should be `/`

k88hudson:
@thisandagain can you take a look at this approach and tell me what you think?

k88hudson:
This demonstrates the strategy on the cordova side of things https://github.com/k88hudson/webmaker-app-cordova/commit/44708cb7e65f3c68c9265d67786ba886a866a9e0

views/add/index.js
```diff
@@ -25,7 +25,7 @@ module.exports = view.extend({
         ],
         goBack: function (e) {
             e.preventDefault();
-            global.history.back();
+            this.page(this.$data.back);
```
thisandagain: :+1:

views/detail/index.js
```diff
@@ -6,7 +6,7 @@ module.exports = view.extend({
     id: 'detail',
     template: require('./index.html'),
     data: {
-        back: true,
+        back: '/templates',
```
thisandagain: This is going to cause an old issue to resurface which is that users will always return to `/templates` regardless of the starting position. As per spec, a user can enter the detail view through discovery or through templates.

thisandagain:
Really solid and much less of a change than I expected. Awesome @k88hudson :+1:

views/detail/index.js
k88hudson: true. I  was originally exploring a hashbang strategy but I can actually put this back in now

k88hudson:
@thisandagain re-R?

thisandagain:
R+

PR #719, opened 2015-01-15
Adding gallery toggle

### LINKED ISSUE #694: Allow users to opt in/out of being a featured app
Sounds like we need something like a checkbox during the publish step for a user to decide whether they want their app featured or not.

This will also require another param on the app objects. I'm thinking `isDiscoverable` (boolean).

gvn:
Resolves #694 

R? @k88hudson

### DIFF
--- a/components/toggle/index.less
+++ b/components/toggle/index.less
@@ -1,0 +1,46 @@
+.toggle {
+    width: 40px;
+
+    .wrapper {
+        position: relative;
+        height: 28px;
+
+        &.checked {
+            label {
+                border: 2px solid @turquoise;
+                left: 100%;
+                transform: translate(-100%, 0);
+            }
+
+            .track {
+                background-color: @turquoise;
+            }
+        }
+    }
+
+    label {
+        background: @white;
+        border-radius: 50%;
+        border: 2px solid @mediumGrey;
+        width: 28px;
+        height: 28px;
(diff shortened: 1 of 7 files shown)


components/toggle/index.less
```diff
@@ -0,0 +1,46 @@
+.toggle {
+    width: 40px;
+
```
k88hudson: you need to add `inline-block` for this to work on webview 4.2 properly

views/share/index.html
```diff
@@ -22,6 +22,17 @@
             </div>
         </div>
     </li>
+    <li>
+        <div class="cell">
+            <div class="left">
+                <span class="title">Webmaker</span>
+                <span class="text-small">Show in Discover Gallery</span>
```
k88hudson: this needs to be `@mediumGrey`

k88hudson:
This is pretty awesome! A couple of small things to fix up. 

One major issue is that this seems to not work/only work sometimes on android 4.2 (click events only work about 10% of time). I'm not sure why :(

k88hudson:
I fixed the UI stuff and rebased in https://github.com/mozilla/webmaker-app/pull/739

We can sort out why stuff is not working on Android in a future sprint

k88hudson:
Also, the toggler is super awesome looking, thanks :smile:

PR #720, opened 2015-01-15
Adds `npm start` script to package

gesa:
(also uses relative path to gulp so it doesn't have to be globally installed)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,8 @@
     "url": "https://github.com/mozillafordevelopment/webmaker-app"
   },
   "scripts": {
-    "test": "gulp test"
+    "test": "gulp test",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -6,7 +6,8 @@
     "url": "https://github.com/mozillafordevelopment/webmaker-app"
   },
   "scripts": {
-    "test": "gulp test"
+    "test": "node_modules/gulp/bin/gulp.js test",
+    "start": "node_modules/gulp/bin/gulp.js dev"
```
k88hudson: `gulp test` and `gulp dev` are enough

k88hudson:
:+1:

PR #721, opened 2015-01-15
only show new non-guest apps whose makers allow featuring

### LINKED ISSUE #700: Filter out guest apps from "Newest" in discovery
Let's not show guest created apps.

Depends on #694

gvn:
Resolves #700


gvn:
R? @k88hudson

Issue #713, opened 2015-01-14
App settings view should have Save button


k88hudson:
To get to it: edit an app, click the Title in the edit view

Current implementation:
![image](https://cloud.githubusercontent.com/assets/1455535/5744357/386f6eec-9bee-11e4-905b-78da6d2f0096.png)

Should have a save button on the right instead, like this:
![image](https://cloud.githubusercontent.com/assets/1455535/5744372/5a5b75f0-9bee-11e4-97d8-ae4eeea18700.png)

k88hudson:
Done!

PR #724, opened 2015-01-16
#713  App settings view should have Save button

gesa:


views/make/index.js
```diff
@@ -35,6 +35,10 @@ module.exports = view.extend({
         goBack: function (e) {
             e.preventDefault();
             if (this.$data.mode === 'settings') {
+                app.update({
```
k88hudson: this will actually add new properties to the remote data model, which is not what we want. i think it might be better to save the old values and re-apply them on cancel  instead, or else create some stand-alone properties on `data` and use those for the settings view instead. what do you think?

k88hudson:
Also thanks for grabbing this one ^_^

views/make/index.js
gesa: So I thought about that, but wasn't really sure the best way to scope a variable to multiple methods. Open to suggestions!

k88hudson: you can set it on `this.$data`, and it will be available in all methods

gesa:
Okay so I made those changes like you asked. My only discomfort with this approach is that it's saving changes as you go, and then reverting them when you hit cancel. This may be an edge case, but I keep thinking about a person making changes on the settings page, then something somewhere failing and they have to, for example, reboot/refresh. They never got a chance to hit cancel because we were disingenuous about what the buttons meant.

...maybe i'm over thinking this.

gesa:
Anyway, it works ^_^

k88hudson:
@gesa you're right, that edge case *is* a possibility, although I feel like it's small enough/not high cost enough not to matter. Thanks for the patch :dancer:

k88hudson:
Although... even as I pressed "comment" on that, I kinda feel like you're right. If it *does* cause issues we can revisit

PR #752, opened 2015-01-19
[Fix #745] Docs and class for displaying apps

### LINKED ISSUE #745: Implement "display" component and view
Right now links in discovery don't actually link to anything. That's because we don't have a view that just displays an app, without making it editable.

It needs to be implemented, and linked to in discovery.

You may be able to just reuse code from the publish view

k88hudson:


k88hudson:
@gvn R?

views/make/edit-play.html
```diff
@@ -1,16 +0,0 @@
-<a class="add"
```
k88hudson: This was dead code

k88hudson:
actually @gvn is away. @thisandagain R?

k88hudson:
I merged this with another PR

Issue #776, opened 2015-01-21
localize unlocalized strings


alicoding:
Hmm sound confusing, but this ticket is to track any strings on the app that is not localize or basically find any hard-coded strings and localize them!

alicoding:
**Page**: 
discover
**Strings**:
Featured
Newest
Screenshot:
![screenshot 2015-01-21 18 40 31](https://cloud.githubusercontent.com/assets/4368149/5847623/1d21de4a-a19d-11e4-8ece-9a020c178aed.png)

alicoding:
**Page**: 
profile
**Strings**:
Make an App
Screenshot:
 
![screenshot 2015-01-21 18 42 06](https://cloud.githubusercontent.com/assets/4368149/5847634/4cc3e22e-a19d-11e4-8c63-9b99a82b0ac2.png)

alicoding:
**Page**: 
make
**Strings**:
Newest
Oldest
Select all
No Data available
Screenshot:
 
![screenshot 2015-01-21 18 44 07](https://cloud.githubusercontent.com/assets/4368149/5847646/7fae5016-a19d-11e4-8a4f-0166198707a7.png)

alicoding:
**Page**: 
make
**Strings**:
* Newest
* Oldest
* Select all
* Make & Share the Web


Screenshot:
 ![screenshot 2015-01-21 18 50 13](https://cloud.githubusercontent.com/assets/4368149/5847725/65238cba-a19e-11e4-8cab-a86fb6809d8c.png)

Note: This is because the key does not match in the templating e.g. "&" vs "&amp;"

alicoding:
**Page**: 
make
**Strings**:
* Submit
* Counter


Screenshot:
![screenshot 2015-01-21 18 55 32](https://cloud.githubusercontent.com/assets/4368149/5847776/35d06be4-a19f-11e4-84b2-a81db1b1ff8b.png)

alicoding:
**Page**: 
make / for counter
**Strings**:
* Button color
* Button text


Screenshot:
![screenshot 2015-01-21 18 56 50](https://cloud.githubusercontent.com/assets/4368149/5847792/5a32a538-a19f-11e4-9c27-ded4780333f6.png)

alicoding:
**Page**: 
make / for counter
**Strings**:
* Title Text Color
* Initial Number
* same as minimum number
* Title
* Your title goes here
* Maximum number
* Minimum number
* Increment by

[...]

PR #792, opened 2015-01-24
fix issue #776 - localize unlocalized strings

alicoding:


alicoding:
@k88hudson can you r? please.

Also, I don't know what exactly is failing on travis right now... I can't see why.

lib/storage.js
```diff
@@ -46,6 +46,17 @@ Store.prototype._createAppRef = function (id, fbRef, data) {
...
+        var localBlock = block;
+        if(block.type === "text") {
+            localBlock.attributes.innerHTML.value = i18n.get(localBlock.attributes.innerHTML.value) || localBlock.attributes.innerHTML.value;
+        } else if(block.type === "phone") {
+            localBlock.attributes.innerHTML.value = i18n.get(localBlock.attributes.innerHTML.value) || localBlock.attributes.innerHTML.value;
+        } else if(block.type === "sms") {
+            localBlock.attributes.innerHTML.value = i18n.get(localBlock.attributes.innerHTML.value) || localBlock.attributes.innerHTML.value;
+        } else if(block.type === "submit") {
+            localBlock.attributes.innerHTML.value = i18n.get(localBlock.attributes.innerHTML.value) || localBlock.attributes.innerHTML.value;
+        }
```
k88hudson:
```js
if (['text', 'phone', 'sms', 'submit'].indexOf(block.type) !== -1) {
    clone.attributes.innerHTML.value =  i18n.get(block.attributes.innerHTML.value);
}
```
you don't need the `||` since it is taken care of in i18n.js https://github.com/mozilla/webmaker-app/blob/master/lib/i18n.js#L65

lib/storage.js
```diff
@@ -46,6 +46,17 @@ Store.prototype._createAppRef = function (id, fbRef, data) {
 
     function newBlock(blockId) {
         var block = blocks[blockId];
+        var localBlock = block;
```
k88hudson: you should do 
```
if (!block) return;
var clone = clone(block);
```

or else you will be directly modifying the original `blocks` object

k88hudson:
If you run `gulp test` you will see the lint errors. Mostly related to spacing for `if` blocks

Awesome start :)

alicoding:
@k88hudson I've another commit and with test failing fix as well :)

lib/storage.js
```diff
@@ -46,7 +46,13 @@ Store.prototype._createAppRef = function (id, fbRef, data) {
...
     function newBlock(blockId) {
         var block = blocks[blockId];
-        return block && clone(block);
+        if (!block) return;
+        var clonedBlock = clone(block);
+        if (['text', 'phone', 'sms', 'submit'].indexOf(block.type) !== -1) {
+            clonedBlock.attributes.innerHTML.value =
+              i18n.get(clonedBlock.attributes.innerHTML.value);
+        }
+        return clonedBlock && clone(clonedBlock);
```
k88hudson: you dont have to reclone here

alicoding: So just return `clonedBlock` right?

PR #796, opened 2015-01-26
adding sideMenu component

gvn:
This is just adding the basic side menu, which can be shown and hidden by clicking the hamburger or outside the menu respectively.

R? @thisandagain

### DIFF
--- a/components/navigationBar/index.less
+++ b/components/navigationBar/index.less
@@ -62,7 +62,7 @@
         color: @highlight;
         border: none;
         outline: none;
-        padding: 0px 20px 0;
+        padding: 20px;
         line-height: (@header-height - 10s);
         text-decoration: none;
         font-size: 13px;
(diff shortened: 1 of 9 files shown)


static/images/menu.svg
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
k88hudson: what is this for?

k88hudson:
A couple of things i noticed testing on-device on 4.2:
* animation is a bit choppier than usual, I'm not sure why. could just be vuejs/whatever else etc. so wouldn't worry about it too much
* the click area for the button needs to be larger

Generally looking super nice though :dancer:

static/images/menu.svg
gvn: Hah, not sure. Just using the assets that Ricardo sent me.

I could actually just remove these SVGs since we are using Font Awesome and they are all FA icons.

gvn:
@k88hudson Do you mean the hamburger needs to be bigger, or...?

k88hudson:
@gvn nah you just need to make the tappable area around the hamburger to be bigger

gvn:
K! :hamburger:

Issue #797, opened 2015-01-26
"Make and share the web" string needs to be rendered as HTML


k88hudson:
This is on `localhost:8080/sign-in`:

![image](https://cloud.githubusercontent.com/assets/1455535/5909117/9279b17c-a57b-11e4-9ec3-eb424f41e8dc.png)

k88hudson:
The directive needed for this is `v-bind-html`

k88hudson:
@pippinlee is going to take this one

PR #799, opened 2015-01-27
Fix html rendering for sign-in #797

pippinlee:
preview preview preview

### DIFF
--- a/_docs/localization/README.md
+++ b/_docs/localization/README.md
@@ -54,3 +54,17 @@ Then, we add the strings to [locale/en_US/mobile-appmaker.json](https://github.c
     }
 }
 ```
+
+## Custom localization directive
+
+Along with the [built-in](http://vuejs.org/api/directives.html) Vue directives is a custom `v-bind-html` directive that is defined in `lib/i18n.js`.
+
+We can see an example of `v-bind-html` in `views/sign-in/index.html`:
+
+```html
+<h1><span v-bind-html="'Make and share the web'"></span></h1>
+```
+
+Here the `v-bind-html` directive will automatically pass the HTML string of "Make and share the web" as the key through the `i18n` filter, which will provide the correct localized content.
+
+Whereas using the `{{ }}` syntax allows you to escape HTML, using `v-bind-html` doesn't escape HTML.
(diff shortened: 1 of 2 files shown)


thisandagain:
:smile: Thanks @pippinlee !

thisandagain:
@alicoding R?

alicoding:
Thanks for the fix @pippinlee but can I ask you to change it to this if you don't mind? 

``` html
<h1><span v-bind-html="'Make and share the web' | i18n"></span></h1>
```

pippinlee:
Hey @alicoding, speaking with @k88hudson, she explained that we likely don't need to include the `| i18n` filter because `v-bind-html` already runs the keys through `i18n`.

Have updated the PR without the `i18n` filter, but will happily change it back if you think we should keep it.

alicoding:
I'm okay with both keeping it or not, but I prefe we do which will help us know that these are being localized (though just v-bind-html still can tell the same) but I guess it just nice to see it when we need to look through? So ill let you decide here :)

pippinlee:
Good ideas!

Agree that using `v-bind-html` and explicitly including the `i18n` filter helps with understanding what is happening. Balancing that with #803, it may make sense to just change the directive name so that this is clearer down the road.

For this PR let's do without it, and then what I've done is updated this PR with more details in the docs. It should hopefully help clarify the use of this custom directive.

If I'm missing anything let me know 👍

alicoding:
I like this! Lets land this one then update the directive name to be more explicit about `i18n` as suggested in the next PR. r+ [[unflagged]](http://plzkthx.herokuapp.com)

thisandagain:
:+1:

alicoding:
Not sure how to unflag my request here on plzthx hehe @thisandagain sorry for the noise here but just trying to get that off.

davidascher:
@alicoding: in plzkthx if your'e signed in you should have a trash icon
next to the mention.

alicoding:
@davidascher It is now gone, but this one was the only one without trash icon :(

davidascher:
i'll probably kill plzkthx.  Let me know if  you find this kind of bug on
the build.webmaker.org site when logged in.

alicoding:
@davidascher sure will do. Right now they are both gone on build and plzthnx

PR #812, opened 2015-01-27
Adding modal prompt and enabling app deletion via sidebar

gvn:


gvn:
R? @k88hudson or @thisandagain

thisandagain:
You write :sparkles:-ly CSS. Thanks for breaking this up into smaller PRs. R+

k88hudson:
a bit broken on 4.2:
image

otherwise beautiful, although i would personally remove that rule or make it `@lightGrey` instead of `#CCC`

PR #850, opened 2015-01-29
Fixes #830: Counter needs to display full width

### LINKED ISSUE #830: Counter needs to display full width
Currently, the counter is not line breaking, causing it to look squished and broken.

screenshot 2015-01-28 13 05 18

gesa:


thisandagain:
R+!

PR #851, opened 2015-01-29
adding inlineEditor component

gvn:
R? @k88hudson

### DIFF
--- a/components/inlineEditor/index.html
+++ b/components/inlineEditor/index.html
@@ -1,0 +1,19 @@
+<div class="wrapper" v-class="editing: isEditMode">
+    <div class="button-wrapper">
+        <button v-on="click: startEditing" class="btn-edit"><i class="fa fa-pencil"></i></button>
+    </div>
+
+    <div class="editor-wrapper">
+        <div class="block-wrapper">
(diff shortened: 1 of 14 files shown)


k88hudson:
This is AWESOME way to go. A few little things:

* transition on background fading to black feels a tad slow to me, maybe speed up by 100ms.
* unfortunately css is busted on the buttons on 4.2. Seems to be a weird thing with flexbox where you need to declare children as block (inline/inline-block won't work). adding `display: block; text-align: center` should do the job
* the designs have a 3 or 4px border radius on the bottom left and right corners
* moving elements is too jarring, needs a transition. happy to file as a new ticket

screenshot_2015-02-02-11-17-45

gvn:
Thanks! 

I took care of the first 3 points, but yeah, if we can file another ticket for refining the post-edit transitions that'd be awesome. :metal: 

R? @k88hudson

thisandagain:
Feel free to drop the animation / transition tickets into the UI improvements milestone. :smile:

k88hudson:
Nice looks good R+

gvn:
Filed follow on ticket #875

PR #859, opened 2015-01-31
Fixes #833:  Add animation to icon chooser

### LINKED ISSUE #833: Add animation to icon chooser
We should add an animation to the icon chooser, it feels like a swipe gesture so we should animate for that effect.

screenshot 2015-01-28 13 11 29

gesa:
It's kind of hackey, but it's sort of the easiest best solution rn.

Originally I thought I was going to use the image names as css classes, but that idea ultimately didn't pan out. That said, I'm leaving the conversion of `app.iconImage` from path to name because I do think it's a more elegant solution, and leaves `app.iconImage` as a more re-usable piece of the data.


gesa:
Those empty groups are pruned in a later commit.

gesa:
Sorry about that @k88hudson, I got a little overzealous when I reverted a thing. Should be resolved now.

k88hudson:
I'm seeing a lot of jank during the transition esp. in chrome and android 4.4, and the image that's transitioning in just pops in (no transition). Maybe this is because they need to be preloaded and/or background transitions just don't perform well?

What if you just did something like render all the images with a repeater and set them to  `v-show: $index === currentIconIndex` and add the `v-transition` property?

then you could add like 
```
transition: transform 0.6s ease;
transform: translateX(0);
&.v-enter {
    transform: translateX(-100%);
}
&.v-leave {
    transform: translateX(100%);
}
```
to the `.icon-image` class instead

k88hudson:
ooor do something totally different lol. whatever works

gesa:
So I totally dig this, and have been playing around with it, but here's the kicker and the reason I didn't try this in the first place: this only has the icons moving in one direction—specifically speaking, it only looks correct when you click the right arrow. It looks like it's moving backwards when you click the left arrow. I've tried a good bit of mental gymnastics, but can't figure out a good way to avoid that.

k88hudson:
hm, good point. maybe try changing the class name dynamically to `transitionForward` and `transitionBack` or something?

even if you can get it to work with just a forward transition, i think that will be an improvement over not having one at all. we can always file a follow up ticket to make further adjustments, but i think it's import to at least get it to be smooth

k88hudson:
also HOLY crap it shouldn't be this hard to do transitions on the web. seriouslyyyyy

gesa:
@k88hudson this is ready as long as you only hit the left arrow.

gesa:
(i mean, the right arrow works too, it just looks... confusing.)

views/make/index.js
```diff
@@ -186,6 +186,9 @@ module.exports = view.extend({
             onValue(app.data);
             self.$data.originalIconImage = self.$data.app.iconImage;
             self.$data.originalIconColor = self.$data.app.iconColor;
+            self.$data.currentIconIndex = self.$data.iconImages
```
k88hudson: you can just do `Math.max(iconImages.indexOf(app.iconImage), 0)` here

k88hudson:
So, i can't for the life of my figure out why, but for some reason this is not working in Chrome. do you get the same error?

gesa:
it's rendering `{{icon}}` as `%7B%7Bicon%7D%7D`

gesa:
(as opposed to, for example, rendering `{{icon}}` as  `/images/icons/howto.svg`)

k88hudson:
@gesa I'm pretty sure this is a bug with v-transition. how would you feel about punting this?

views/make/settings.html
```diff
@@ -10,7 +10,8 @@
     </div>
     <div class="icon-container">
         <div style="background: {{app.iconColor}}" class="app-icon">
-          <div class="icon-image" style="background-image:url('{{app.iconImage}}')"></div>
+
+          <div class="icon-image" id="icon-{{$index}}" v-repeat="icon: iconImages" v-show="$index === currentIconIndex" v-transition style="background-image:url('{{icon}}')"></div>
```
BigBlueHat: Maybe give `v-style="background-image: 'url(' + {{icon}} + ')'"` (untested...just FYI :smile_cat:)

For reference: http://vuejs.org/api/directives.html#v-style

gesa:
Okay, look. It works everywhere but Chrome, this *has* to be a Vue.js bug.

For posterity's sake:
at <https://github.com/gesa/webmaker-app/blob/833/animate-icons/views/make/settings.html#L13>, Vue evaluates `$index` correctly on load. It evaluates `currentIconIndex` correctly on load. Yet for some reason, it evaluates  `v-show="$index === currentIconIndex"` on load as truthy no matter what. In `onClick` functions, it evaluates `$index === currentIconIndex` correctly.

thisandagain:
Closing for now. Can reopen after this sprint. Thanks @gesa !

PR #901, opened 2015-02-04
Fixes #900:  Development server can't be accessed from local network

### LINKED ISSUE #900: Development server can't be accessed from local network
This makes testing on actual devices incredibly challenging!

According to [this](https://github.com/schickling/gulp-webserver#why-cant-i-reach-the-server-from-the-network) this can be resolved by adding  `host: '0.0.0.0',` to https://github.com/mozilla/webmaker-app/blob/master/gulpfile.js#L74-L78. (I tested this theory, it's true.)

gesa:
I should note that localhost:8080 will still work. Just now [your local IP]:8080 will as well.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,6 +72,7 @@ gulp.task('watch', ['build'], function () {
 gulp.task('dev', ['watch'], function() {
   return gulp.src('build')
     .pipe(webserver({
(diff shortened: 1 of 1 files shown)


thisandagain:
Nice! Thanks @gesa ! R+

PR #923, opened 2015-02-06
Fixes #868: Implement new app detail view

### LINKED ISSUE #868: Implement new app detail view
Design: https://redpen.io/wy5af842444b6a2667
( Design ticket: https://github.com/mozilla/webmaker-app/issues/839 )

Note that:

* this should replace the current `detail` view
* the edit app button and app data tab should only appear if the user owns the app

gesa:


components/navigationBar/index.html
```diff
@@ -1,12 +1,12 @@
...
-    <a v-if="typeof back === 'string'" class="back nav-btn" href="{{back}}">&lt;</a>
+    <button v-if="back === true" class="back nav-btn" v-on="click: goBack"><span class="icon ion-android-arrow-back"></span></button>
+    <a v-if="typeof back === 'string'" class="back nav-btn" href="{{back}}"><span class="icon ion-android-arrow-back"></span></a>
 
     <button v-if="cancel === true" v-on="click: goBack" class="nav-btn">{{'Cancel' | i18n}}</button>
     <a v-if="typeof cancel === 'string'" href="{{cancel}}" class="nav-btn">{{'Cancel' | i18n}}</a>
 </div>
 
-<h1>{{title | i18n}}</h1>
+<h1>{{app.name}}</h1>
```
k88hudson: This is used for all views, so changing this will affect them too -- instead, what you want to do in share/index.js is set the `title` property around here https://github.com/mozilla/webmaker-app/blob/master/views/share/index.js#L80.

So like
```
self.$data.title = app.name
```

Is it weird that child components like navigationBar automatically inherit their parent's `$data` properties? Definitely. Is it the way vue.js 10.6 works? Yes.

k88hudson: If you really think this is gross, which it is, you can alternatively change this line in share/index.html https://github.com/mozilla/webmaker-app/blob/master/views/share/index.html#L1
```
<nav v-component="navigationBar" class="nav-default" v-with="title: app.name"></nav>
```

That will explicitly pass the `app.name` value along as the `title` property in navigationBar.

k88hudson:
* "Show this app in the Discovery Gallery should be 14px and `@darkGrey`
* Open my app button should be 14px, remove inset box shadow, box shadow on the bottom should be #518AC2, vertical padding should be `15px` on top and bottom
* When not pressed, the `top` value should be `0`, the pressed value should be `2px`
* Active state should have bg color #5C9CDB

k88hudson:
This is an awesome start, can't wait to see this on master :dancer:

gesa:
just one note on `.btn-press`: when i stole that code from webmaker.org, i was thinking about dropping the hover state. but then i remembered that my sister just got this phone (the note, maybe?) that totally supports hover states, (and apparently that's been a thing [at least since 2012](https://en.wikipedia.org/wiki/Samsung_Galaxy_Note_II)) so i left it in. thoughts @k88hudson ?

k88hudson:
@gesa I wouldn't bother with hover states for now

gesa:
Could you take a look @k88hudson? I have some TODOs in there that need either another set of eyes, or to be declared out of scope.

static/styles/common.less
```diff
@@ -50,6 +50,16 @@ body, html {
         transition: opacity 0.2s linear;
     }
 }
+
+
+// TODO maybe find/create a more appropriate less file for this
+.app-preview {
+  border-radius: 4px;
+  background: @white;
+  overflow: scroll;
```
k88hudson: you can actually remove this now

k88hudson: the overflow: scroll that is

k88hudson:
Ok this is pretty much good to go, can you just fix that one nit and hide the my app/app data as well as the show in gallery toggle for the non-admin view?

PR #951, opened 2015-02-07
button restyling

### LINKED ISSUE #818: Implement New Buttons
Let's do some iterations of how our buttons styles could look like.

Pomax:
fixes #818 (well, implements)

- button master class in button.less adapted to @vazquez's CSS
- changed the delete/cancel buttons to use classes from buttons.less
- current uses some "magic" colours (e.g. currently suffering from "there is no reason for the same hex codes to mean the same thing"), so those might need to be formalised for this PR is allowed through.

main button:
screenshot 64

share dialog:
screenshot 63

delete dialog:
screenshot 66


thisandagain:
Looks like the CSS styles in @vazquez 's codepen don't quite match his mockups. Here is a quick revision that should get this closer: https://gist.github.com/thisandagain/a6f3f9c3f3678611d34a

thisandagain:
One other issue I'm seeing is that the latest menu styles from @gvn are being affected by these changes too:
screen shot 2015-02-06 at 6 23 16 pm

vazquez:
I have updated the Codepen to include the delete and cancel buttons: http://codepen.io/rvazquez/pen/gbGzye

thisandagain:
@vazquez can you check your sizing? Looks almost like these are at 150% scale.

xmatthewx:
Text on the buttons feels like it's a pixel or two above center.

k88hudson:
@Pomax @vazquez @thisandagain @gesa let's coordinate on this early next week :+1:

vazquez:
The authoritative buttons prototype

http://codepen.io/rvazquez/pen/gbGzye?editors=110

I am working on the Submit button right now, but I think this is enough to get started on the other buttons.

k88hudson:
@Pomax can you also disable the weird background on buttons? 
see http://davidwalsh.name/remove-mobile-firefox-button-gradient

gesa:
I'm going to use these classes in #923, and not introduce my own.

Pomax:
this also weeds a few of the btn-... classes to make sure .btn is the base class and not redefined everywhere. screenshots incoming for what the effect of this has been.

vazquez:
The button prototype now includes Submit button: http://codepen.io/rvazquez/pen/gbGzye?editors=110

static/styles/buttons.less
```diff
@@ -1,77 +1,78 @@
+@btnInsetShadow: inset 0px 2px 5px rgba(0,0,0,0.2);
```
thisandagain: :+1:

static/styles/buttons.less
```diff
@@ -1,77 +1,78 @@
...
-    display: inline-flex;
-    align-items: center;
-    justify-content: center;
+    font-weight: 400;  
+    color: white;
+    line-height: 1;
+    border-radius: 4px;
+    border: medium none;
+    outline: medium none;
+    cursor: pointer;
```
thisandagain: Probably not needed, but ok as it makes testing in-browser a little more natural.

thisandagain:
:cat2: :cake: :boom: R+

static/styles/buttons.less
vazquez: +1 figured that it wouldn't hurt :)

PR #973, opened 2015-02-10
restyling segmented control (aka "switch")

### LINKED ISSUE #828: Implement segmented control
Perhaps a tab control would be better. Segmented control is more iOS-y. Android design patterns favour tab controls.

gvn:
Resolves #828 

R? @thisandagain

### DIFF
--- a/components/switch/index.less
+++ b/components/switch/index.less
@@ -2,9 +2,9 @@
     @border-radius: 30px;
     @height: 50px;
     width: 100%;
-    height: @height;
     user-select:none;
     -moz-user-select:none;
+    text-transform: uppercase;
 
     input[type=checkbox] {
         display: none;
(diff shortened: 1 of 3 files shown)


components/switch/index.less
```diff
@@ -14,12 +14,11 @@
         display: block;
         overflow: hidden;
         cursor: pointer;
-        border: 1px solid @turquoise !important;
         background-color: @turquoise !important;
```
thisandagain: Can we use `@highlight` here?

components/switch/index.less
```diff
@@ -38,37 +37,24 @@
                 text-align: center;
                 position: relative;
                 z-index: 2;
-                transition: all 0.2s ease-in 0.0s;
+                color: #005d8e;
```
thisandagain: We should add a color spec for this as we are going to be using it for the `done` button too.

thisandagain:
R+ :floppy_disk:

PR #979, opened 2015-02-10
removed the .attached handling for tabBar

### LINKED ISSUE #978: tabBar flex styling gets set to "block" on blur
this happens in components/tabBar/index.js, in the attached function

Pomax:
fixes #978, removes the event handling that was left over from an older component, messing up the flex layout by overriding the flex layouts as "block" instead..


gvn:
R+

PR #980, opened 2015-02-10
fixing image display in inline editor

gvn:


gvn:
NOBUG

R? @k88hudson

k88hudson:
yay!

PR #983, opened 2015-02-10
Fix issue #896 - Create a Brick for Links

### LINKED ISSUE #896: Create a Brick for Links
All users (99%)  want to be able to share links in their app. 
They want the link to be either : 
- Internal -> will link to bottom of app or up (avoiding scrolling)
- External

alicoding:
Closes #896

### DIFF
--- a/blocks/link/index.js
+++ b/blocks/link/index.js
@@ -1,0 +1,22 @@
+module.exports = {
+    className: 'link',
+    template: require('./index.html'),
+    data: {
+        name: 'Link',
+        icon: 'images/blocks_link.svg',
+        attributes: {
+            link: {
+                label: 'URL',
+                type: 'input',
+                value: '',
+                placeholder: 'https://mozilla.org'
+            },
+            text: {
+              label: 'Link Text',
+              type: 'input',
+              value: '',
+              placeholder: 'Mozilla'
+            }
+        }
+    }
+};
--- a/blocks/link/index.html
+++ b/blocks/link/index.html
@@ -1,0 +1,1 @@
+<a href="{{attributes.link.value}}">{{attributes.text.value}}</a>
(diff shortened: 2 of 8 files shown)


alicoding:
@k88hudson can you R? please

blocks/link/index.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+    className: 'link',
+    template: require('./index.html'),
+    data: {
+        name: 'Link',
+        icon: 'images/blocks_link.svg',
+        attributes: {
+            link: {
+                label: 'Link Text',
```
k88hudson: This should be URL

blocks/link/index.js
```diff
@@ -0,0 +1,22 @@
...
+                type: 'input',
+                value: '',
+                placeholder: 'https://mozilla.org'
+            },
+            text: {
+              label: 'URL',
```
k88hudson: This should be Link Text

k88hudson:
@alicoding mostly looks good just need to switch those labels

alicoding:
@k88hudson R?

thisandagain:
On Android / Chrome when you add a link, it seems to get obscured by neighboring elements. Most likely this is because it needs to be explicitly defined as a `block` level element.

screen shot 2015-02-10 at 8 05 51 pm

alicoding:
I actually forgot to add padding to the block so it just happen to look empty when the user didn't have any text in there at all.

A question though @thisandagain should we allow empty string on both input at all? Should we make this as required field?

alicoding:
screen shot 2015-02-10 at 23 28 14
screen shot 2015-02-10 at 23 28 17

thisandagain:
R+ :baby_chick: :computer: :zap:

thisandagain:
Thanks @alicoding !

PR #991, opened 2015-02-10
cancel => back label replacement

### LINKED ISSUE #935: Edit Title - Cancel is only option
Android 4.4.4 

Click pencil to edit title. On the next view, `Cancel` is the only way out. It saves rather than cancels changes. There is no Save button. 

Also, it's slightly jarring to see all the colors and icon when I thought I was only going to edit the app name. But, that's a bigger design task than the bug above.

screenshot 2015-02-06 17 38 44

Pomax:
fixes #935 by replacing the incorrect label "Cancel" on the make settings view with the far more accurate "Back" label


thisandagain:
As per the `navigationBar` component we should be using the back icon to represent this rather than the text "Back". In addition, to fully resolve this issue we'll want to add a "√" button to confirm the changes to the top right of the view.

Pomax:
closing this PR, will refile with the better solution (keeping cancel the same but adding a "done" button in the upper right to homogenise with the other views)

PR #998, opened 2015-02-11
color picker always  hides on 'back' now

### LINKED ISSUE #707: Android device back button does not work on all pages
Page: choosing a color. 
Clicks on device back button : does not do anything 
Clicks on app back button: does not go back or quit the app

Almost each time we clicked Android Device back button, the app broke. 

==> Need to check back and save for each page.
It works sometimes on some pages, but not for the majority

### LINKED ISSUE #997: color picker should always go back to the block that opened it
Or, rephrased, the color picker should close itself on the "back" button in Android.

Pomax:
Fixes #997, fixes #707, by intercepting the hardware back button (android, etc) using phonegaps "backbutton" event handler.

### DIFF
--- a/components/colorPicker/index.js
+++ b/components/colorPicker/index.js
@@ -10,7 +10,8 @@ module.exports = {
             this.show = true;
             this.originalColor = event;
             this.cancelIntercept = this.onCancel.bind(this);
-            document.addEventListener("backbutton", this.cancelIntercept, true);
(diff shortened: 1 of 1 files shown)


Pomax:
error on "exports.js", but that seems to be from a different patch...

thisandagain:
:smile: R+

PR #1012, opened 2015-02-11
Updating blue color variable

### LINKED ISSUE #995: Update blue colour variable one last time
We need to update the blue colour variable one last time. We are making the colour a bit brighter now that we are using the blue in headers. The blue now needs to play nice with elements like the universal done button.

vazquez:
Fixes #995

### DIFF
--- a/styles/colors.less
+++ b/styles/colors.less
@@ -2,7 +2,7 @@
 @black: #3B4E5B;
 @white: #FFF;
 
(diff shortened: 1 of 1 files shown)


vazquez:
R? @thisandagain [flagged!]

thisandagain:
R+ :boat: :anchor: :icecream:

PR #1014, opened 2015-02-11
getting rid of all tap highlight colors

### LINKED ISSUE #1002: Hide web-y highlight on "down" state of segmented control
When you press on the segmented control, you get a very web-y highlight of the entire element. This can be seen in both the `discover` and `detail` views.

gvn:
Resolves #1002 

R? @thisandagain

### DIFF
--- a/styles/common.less
+++ b/styles/common.less
@@ -25,6 +25,12 @@ body, html {
     background: none;
 }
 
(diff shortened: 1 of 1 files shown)


thisandagain:
:+1: R+

PR #1022, opened 2015-02-11
universal 'save' button styling

Pomax:
note: not hooked up yet, since that's part of a different ticket.

### DIFF
--- a/components/navigationBar/index.less
+++ b/components/navigationBar/index.less
@@ -32,17 +32,19 @@
 
     .save-btn {
         @_btnSize: 30px;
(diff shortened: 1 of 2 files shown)


Pomax:
shows up in the make title edit view and the color picker

alicoding:
I think you need to remove `<button class="btn btn-block" v-on="click: onSave">{{ 'Save' | i18n}}</button>` from `views/block/index.html` too

Pomax:
different issue - this is for implementing the styling only

vazquez:
R+ for styling! :ear:

PR #1024, opened 2015-02-11
UX hooked up

Pomax:
currently fully implemented in the color picker (start a make, edit a text thing, change color)


vazquez:
1 If I remove all text from the input, the done button needs to disable, as we can't allow an empty titled app.

screenshot 2015-02-11 16 20 47


2 When I tap on the Done button, the screen should disclose and I should immediately come back to: 

screenshot 2015-02-11 16 22 36

vazquez:
R+ for style and interactivity :boom: The button behaves as expected.

PR #1037, opened 2015-02-11
Fixes #1009: Previews should not be interactive

### LINKED ISSUE #1009: Previews should not be interactive
Previews are currently somewhat interactive in that you can edit form elements but not submit them. All elements within previews should not register touch / click events.

screen shot 2015-02-10 at 8 37 13 pm

gesa:
The `disableBlocks` function is *probably* not necessary, but I'm not taking any chances with such a multitude of clients.

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -20,6 +20,11 @@ module.exports = Vue.extend({
     data: {},
     methods: {
         model: require('./model')(),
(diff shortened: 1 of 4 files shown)


thisandagain:
R+ :loud_sound: :leaves:

Issue #849, opened 2015-01-29
localize unlocalized strings v2


alicoding:
There seems to be more strings missed or never localized in these files:

`components/modalPrompt/index.html`
`components/publishFooter/index.html`
`components/sideMenu/index.html`

davidascher:
Is it possible to build a travis-triggered test to detect these?

alicoding:
I think it is possible if we can look for a string inside any of the specify tags or DOM in general. So if we don't use `<p>{{ something | i18n}}</p> ` and have it as `<p>something in here</p>` maybe we can safely assume that this is not localized.

thisandagain:
Let's hold off on automating this until after MWC as we'll likely be porting to react which will alter directives / approach.

davidascher:
+1

thisandagain:
Can this be closed @alicoding ?

Issue #1020, opened 2015-02-11
i18n not showing default locale


gvn:
*Platform:* All

**Repro:**
1. Add `{{ 'No Data available' | i18n }}` to HTML
2. Change `No Data available` value in `/locale/en_US/mobile-appmaker.json` to `No data has been stored yet!`

**Expected:** "No data has been stored yet!" shows up in rendered template.
**Actual:** "No Data available" shows up.

alicoding:
I'll take a look into this, but FYI you should change in webmaker-app.json not mobile-appmaker.json

gvn:
I don't see `webmaker-app.json` in the repo...

https://github.com/mozilla/webmaker-app/tree/master/locale/en_US

alicoding:
Oops! I thought I did update both on Transifex and here.

alicoding:
I have no clue right now what is the problem here, but this is what I've narrowed it down to.
The problem is in `gulp/locale.js` where is compiling all locales into one file in `locale/index.js`

So, for some reason if I add that key it won't get compile in gulp, but I did the same in node console and it's working fine.

I'm going to spend some time looking at this again to see what is really the problem, but to unblock (if this is really blocking you) change the key name with number infront it will make it visible for now (Please note this is just a workaround and I will try to find a permanent fix soon).

alicoding:
Took me forever to know what was the problem here!

https://github.com/mozilla/webmaker-app/blob/master/locale/en_US/mobile-appmaker.json#L62-L65 this is why!

alicoding:
Verified that this is not an issue anymore.

gvn:
:+1:

PR #1047, opened 2015-02-12
At least partial of #849: localize unlocalized strings v2

gesa:


alicoding:
Most of these doesn't seem to work here, but not sure if it is related to https://github.com/mozilla/webmaker-app/issues/1020 but I think I'm going to r+ this anyway since we will need to upload this for translation soon. so R+ [[unflagged]](http://plzkthx.herokuapp.com)

PR #1052, opened 2015-02-12
Fix l10n related issue with sidebar and related one

alicoding:
There are certain places within the app that it doesn't seem to pick up the right locale and it always display the defaultLang which we set to `en-US`.


alicoding:
@k88hudson or @thisandagain can you R?

thisandagain:
Just resolve the lint errors. Otherwise looks good.

alicoding:
@thisandagain done, thanks!

PR #1064, opened 2015-02-12
wiring up gallery toggle

### LINKED ISSUE #921: Share Via Webmaker toggle doesn't do anything
Re this control: (on /make/:id/share)

Toggling to the `true` position does give me *visual* confirmation that a make is shared, but upon reloading the page, the toggle is set to the `false` position.

gvn:
Resolves #921 

R? @thisandagain


thisandagain:
Yay! R+ :shaved_ice: :sheep: :shipit:

PR #1040, opened 2015-02-11
Campus safety (and general phone button edit view) fix

Pomax:
Fixes #976 - The phone button used a property called `innerHTML` with the template using `label`. Conflicts arose. This should pacify said conflicts.

PR #1065, opened 2015-02-12
... better phone fix

### LINKED ISSUE #1048: Can't add a Phone block
*OS:* ALL
*Browser:* ALL

**Repro Steps:**
1. Try to add phone block

**Expected:** Can add phone block.
**Actual:** `Uncaught TypeError: Cannot read property 'value' of undefined` – `storage.js:53`

Pomax:
fixes #1048 - turns out the `innerHTML` propery is shared by so many things that changing it to `label` just for the phone brick, in #1040, created way more problems than it solved.


gvn:
R+

thumbsup

PR #1071, opened 2015-02-12
fix gulp publish task string

k88hudson:


gvn:
R+
 [[unflagged]](http://plzkthx.herokuapp.com)

sk8

PR #1098, opened 2015-02-12
changes how to deal with unknown strings

### LINKED ISSUE #1070: Regression: Title doesn't show up (in markup) on detail view
*OS:* ALL
*Browser:* ALL
*Screenshot:* 

**Repro Steps:**
1. In profile OR discovery view, tap any app to see its detail view

**Expected:** Title renders somewhere on page (presumably in that top nav bar)
**Actual:** No title anywhere!

Pomax:
fixes #1070, the original fix was missing a check to see whether or not the to-resolve string actually led to a translation. This patch updates that code path to only return if a translation is found, otherwise it falls through to the fallback code


alicoding:
r+ very good catch O_O

PR #1112, opened 2015-02-12
calculated counter width

Pomax:
ensures that the counter's input is always 100% - twice the buttons.

### DIFF
--- a/blocks/counter/index.less
+++ b/blocks/counter/index.less
@@ -6,6 +6,8 @@
         align-items: stretch;
         justify-content: space-between;
 
+        @buttonWidth: 45px;
+
         button {
             font-size: 36px;
             color: @turquoise;
@@ -13,6 +15,7 @@
             border: 1px solid @lightGrey;
             padding: 0 15px;
             margin: 0;
+            width: @buttonWidth;
 
             &:first-of-type {
                 border-radius: 5px 0 0 5px;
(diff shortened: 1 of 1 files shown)


blocks/counter/index.less
```diff
@@ -29,6 +32,7 @@
             border-right: none;
             text-align: center;
             color: @turquoise;
+            width: ~"calc(100% - " (2*@buttonWidth) ~")";
```
gvn: calc won't work on Android 4.2 :sob:

gvn:
R- [[unflagged]](http://plzkthx.herokuapp.com)

blocks/counter/index.less
k88hudson: :cry:

gvn:
There's still some issues with layout. The +/- aren't vertically aligned.

I'd recommend *possibly* just refactoring the component so that it doesn't use an input since they're difficult to style.

screenshot 2015-02-13 10 45 41

davidascher:
If we expect people to use it to enter large numbers, we may not be able to
avoid an input widget, no?  Or do you mean a custom from-scratch input?

gvn:
We can still avoid it. The input is still what's used underneath, it's just not displayed. With this solution you have a DIV with inner text that is the same as the number contained in the input, so it will have the same constraints and capabilities while being easier to consistently style.

PR #1121, opened 2015-02-13
adding no interwebs view + refactoring missing content warnings

### LINKED ISSUE #885: Implement offline state for discovery mode, sharing
If users *start* in an offline state, they won't have access to the content in /discovery. We can do a better job at caching previous searches in the future, but for now we should at least have a better "can't connect" state for this view.

I'd love to see something that encourages to try a template instead or something similar.

gvn:
Resolves #885


thisandagain:
Awesome start. :smile:

PR #1134, opened 2015-02-13
remove the overlay in template/...

### LINKED ISSUE #1088: Opaque screen can't be dismissed after back button in inline editor
STR (verified, likely android only)

- open promote your business template
- scroll to counter
- hit pencil on counter
- dismiss by hitting Android hardware back button

--> I have a semitransparent overlay that I can't dismiss

Pomax:
fixes #1088

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -50,7 +50,7 @@ module.exports = function (app) {
     page('/template/:id/detail', function (e) {
         app.currentView = 'template-preview';
         app.params = e.params;
-        app.$broadcast("closeShim");
+        app.$broadcast('closeShim');
     });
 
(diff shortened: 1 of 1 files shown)


lib/router.js
```diff
@@ -50,6 +50,7 @@ module.exports = function (app) {
     page('/template/:id/detail', function (e) {
         app.currentView = 'template-preview';
         app.params = e.params;
+        app.$broadcast("closeShim");
```
gvn: Needs single quotes to pass `gulp lint`

gvn:
Just need to fix the double quotes and then R+

PR #1138, opened 2015-02-13
temporary fix for the done button

### LINKED ISSUE #1046: Universal "Save" button should have disabled state
STR:

1. Add text block
2. First state we're seeing here is disabled state
3. Focus on a textarea (input) then it will be enabled state
4. You can save which is fine, but even if you are in an error state you still can save which is not the behaviour I think it suppose to be.

screen shot 2015-02-11 at 19 58 44
screen shot 2015-02-11 at 19 58 50
screen shot 2015-02-11 at 19 58 53

Pomax:
this does **not** fix #1046, but it's going to have to do due to reasons stated in that bug,.

### DIFF
--- a/views/block/index.js
+++ b/views/block/index.js
@@ -22,7 +22,7 @@ module.exports = view.extend({
     components: editorModels,
     data: {
         title: 'Edit',
-        // The following code has been commented off due to 
+        // The following code has been commented off due to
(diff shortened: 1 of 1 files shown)


gvn:
Just some formatting fixes then good to go

PR #1143, opened 2015-02-13
Fix bug with the toucan and can't click to edit

alicoding:
@thisandagain R?


alicoding:
actually there is another bug I found and will fix it here now too 
screen shot 2015-02-13 at 13 09 00

alicoding:
Fixed 
screen shot 2015-02-13 at 13 09 59

PR #1150, opened 2015-02-13
Fixing position of the add content screen with Tucker the Toucan

vazquez:


vazquez:
The toucan was smaller than it needed to be on the Add content screen and the content blocks were not centred.

screenshot 2015-02-13 14 49 43

alicoding:
@thisandagain R?

PR #1155, opened 2015-02-13
Fixes #1079:  icons in the inline editor are too small

### LINKED ISSUE #1079: icons in the inline editor are too small
At least on my nexus 5, the icons feel like they could easily be 2x bigger

gesa:
From #1079, icons about double the size:
screen shot 2015-02-13 at 12 47 51


gvn:
R+!

ohyeah

PR #1166, opened 2015-02-14
Fixes #1097: Text block editor has a resize handle

### LINKED ISSUE #1097: Text block editor has a resize handle
Strangely enough, the resize handle actually works, but

a) that's not an affordance that mobile users will understand
b) the resize handle is way too small.

Suggest we just make the text box not really resizable by hand. (It should grow with large text bodies though).

gesa:
Should solve #1097. Didn't require any additional javascript to work, either. Yay!


k88hudson:
In my experience `contenteditable` is not a great option for text that needs to be persisted since it has a lot of inconsistencies between browsers and can often lead to unexpected characters and tags (e.g.  newlines/breaks, style tags in copied text, etc). 

Maybe try `max-height` and `max-width` instead?

PR #1178, opened 2015-02-17
Updating icon and fixing position of add icon in navigation bar

### LINKED ISSUE #1170: The + button in the top right
The + button the to top right corner to add content to an app seems rather small and way up there in the corner.

Maybe it would be nice to have the buttons the same style as in the Firefox OS contact app and other built in apps. There the button also gets highlighted when you move the cursor over it, which is nice in the simulator and desktop.

vazquez:
Fixes #1170

### DIFF
--- a/components/navigationBar/index.less
+++ b/components/navigationBar/index.less
@@ -93,6 +93,7 @@
         font-size: 13px;
         .ion {
             font-size: 20px;
+            line-height: 1.4;
         }
         &.nav-btn-icon {
             font-size: inherit;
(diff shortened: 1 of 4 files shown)


k88hudson:
Looks good in itself but there are a few problems balancing with the other elements of the header:
* + is misaligned with the pencil
* outer right margin is uneven compared to the outer left margin of the hamburger
image

Also, it looks like the hamburger is too thin/small compared to plus - you might want to bump up the size of that too.

vazquez:
screenshot 2015-02-17 11 34 42

PR #1194, opened 2015-02-18
Change zIndex to 99 while animating (Closes #941)

### LINKED ISSUE #941: When moving, an item should be on top of the others
i.e. If I select the item that's undeneath the clipboard image in the blogger template and hit the up arrow, it moves behind the image, whereas I'd expect (esp. based on Android's Material standard) to move on top.

i.e. set a higher z-index before transition.

errietta:


components/blockList/index.js
```diff
@@ -138,6 +139,8 @@ module.exports = {
                 element.style['-webkit-transform'] = value;
             }
 
+            elTarget.style.zIndex = '99';
```
gesa: The navbar's `z-index` is set to 10, would this work if it was set to 9?

gesa:
:shipit:

PR #1195, opened 2015-02-18
more i18n strings!

gesa:
relates to #849


alicoding:
I'm R+ this for localization work. @thisandagain once this land I'm planning to upload the file to Transifex again since we have two new sets of strings to translate again! :P

thisandagain:
Thanks for updating this @gesa . I didn't notice we had missed these last week. Thanks @alicoding !

PR #1221, opened 2015-02-19
hiding image picker for non image blocks

### LINKED ISSUE #1090: Counter +/- buttons don't work in the editor
STR:
 - open Promote your business template
 - tap on the +/- buttons in the counter bricks and see nothing happen
 - tap "Start building"
 - tap on the +/- buttons in the counter bricks and see nothing happen
 - hit pencil on that button, see the inline editor
 - tap on the +/- buttons in the counter bricks and see nothing happen

At least the last two seem like reasonable ways to set the default value.  Having to go to the property editor seems like unnecessary long path.

gvn:
Fixes #1090

### DIFF
--- a/components/inlineEditor/index.html
+++ b/components/inlineEditor/index.html
@@ -8,7 +8,7 @@
     <div class="editor-wrapper">
         <div class="block-wrapper">
             <content select=":first-child"></content>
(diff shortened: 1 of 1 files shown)


k88hudson:
:space_invader:

PR #1224, opened 2015-02-19
Handle data submission in offline state

### LINKED ISSUE #1133: Handle data submission in offline state
STR:
- go offline (added by @thisandagain)
- choose survey app
- tap start building
- scroll to the bottom
- tap the `Submit` button

Nothing happens (and no data is visible in the app's Data tab)

gesa:
Fixes #1133

### DIFF
--- a/components/publishingStatus/index.js
+++ b/components/publishingStatus/index.js
@@ -5,6 +5,9 @@ var analytics = require('../../lib/analytics');
 module.exports = {
     className: 'publishing-status',
     template: require('./index.html'),
+    data: {
+        isOnline: true
+    },
     ready: function () {
         var self = this;
         self.$data.isPublishing = isPublishing;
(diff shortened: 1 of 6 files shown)


alicoding:
I'm adding `blocked` label for l10n in this file since I can't upload and asking the contributors to translate again frequently.

thisandagain:
:+1:

components/publishingStatus/index.js
```diff
@@ -28,5 +31,13 @@ module.exports = {
                 self.$data.isPublishing = isPublishing = false;
             }, 5000);
         });
+        this.$on('publishOffline', function () {
+            self.$data.isOnline = false;
+            self.$data.isPublishing = true;
+            console.log('bar');
```
alicoding: console.log :P

thisandagain:
I'd prefer to land this (in english) and hold off on updating the Transifex file till after MWC. Thoughts @alicoding ?

alicoding:
@thisandagain if this patch is not going to be merge and ready in an hour then I would say we hold off on updating this on Transifex, but either way I have to update the file on Transifex anyway since we found few unlocalized strings and they have to be localized.

components/publishingStatus/index.js
gesa: Good catch ^_^

gesa:
I fixed that `console.log()`, should be good to go.

components/publishingStatus/index.js
```diff
@@ -5,6 +5,9 @@ var analytics = require('../../lib/analytics');
 module.exports = {
     className: 'publishing-status',
     template: require('./index.html'),
+    data: {
+        isOnline: undefined
```
alicoding: set this to `true` instead of `undefined`.

alicoding:
r+

PR #1241, opened 2015-02-19
Fix lint errors in analytics library

thisandagain:
Will explore apparent race condition in the test suite separately. This should resolve the intermittent `lint` failures.


thisandagain:
@gesa R? [[flagged!]](http://plzkthx.herokuapp.com)

gesa:
R+, :rocket:

PR #1242, opened 2015-02-19
less aggressive saving of app name during edits

### LINKED ISSUE #1163: Editing App title is really difficult
Just tried to remix a couple different apps from the discover tab. When I went to change the title of each app, parts of the title would pop back into the textbox. Particularly the "My" and "App" bookends.

gvn:
Fixes #1163


thisandagain:
R+ :electric_plug:

PR #1249, opened 2015-02-20
Fix #1210 - Bring back nuke my apps

### LINKED ISSUE #1210: bring back the delete-all button, in the super-secret-squirrel-developer menu
This is not a feature that should be in the user menus, but definitely makes sense in the hidden dev menu, because deleting 200 apps one by one is intense! =D

alicoding:
Closes #1210


thisandagain:
R+ :koala:

Issue #956, opened 2015-02-09
Add Brick cancel flow


xmatthewx:
As a user new to Webmaker, I'd like to click a brick in the brick selector, view the edit options, and then cancel to return to the selector (rather than the app). 

I believe this is more intuitive and conducive to learning and experimentation. The jump back to the app requires me to click add again. 

This would require a cancel button in the selector view, which would require some new design.

thisandagain:
Major :+1: to this.

vazquez:
Definitely :+1: The selector screen already has a cancel button, we are just not routing it correctly.

![screenshot 2015-02-19 11 35 16](https://cloud.githubusercontent.com/assets/1119821/6270945/787cc8a0-b82b-11e4-89e5-e2003ea12724.png)

**Action:** When a user taps on "Cancel", we need to send them back to the Add Content screen, instead of the Edit screen.

alicoding:
@thisandagain currently if you click on the editor and pick "Text" block then click "Cancel" it will add the block. Is this an expected behaviour? I think it is weird when I click cancel and it's still adding the block for me?

thisandagain:
@alicoding Yikes. That's a bug. Can you file?

PR #1253, opened 2015-02-20
Fix #1252 - Cancel shouldn't add block to my app

### LINKED ISSUE #1252: Cancel on adding block page shouldn't add a block
STR:

1. Click "Make" tab
2. Choose "Start From Scratch"
3. Tap the "+" sign
4. Choose "Text" block
5. Tap "Cancel" on top left

Expected:
Shouldn't add Text block to my app

What happen:
Text block was added to my app

alicoding:
Closes #1252 

And should also fix https://github.com/mozilla/webmaker-app/issues/956

### DIFF
--- a/components/inlineEditor/index.html
+++ b/components/inlineEditor/index.html
@@ -21,7 +21,7 @@
                <img src="content/icons/duplicate.svg">
             </button>
 
-            <a v-on="click: onEditClick" class="btn btn-flat" href="/make/{{ app.id }}/block/{{ index }}">
+            <a v-on="click: onEditClick" class="btn btn-flat" href="/make/{{ app.id }}/block/{{ index }}/edit">
                 <img src="content/icons/edit.svg">
             </a>
 
--- a/lib/router.js
+++ b/lib/router.js
@@ -60,7 +60,7 @@ module.exports = function (app) {
         app.$broadcast('closeShim');
     });
 
-    page('/make/:id/block/:index', function (e) {
+    page('/make/:id/block/:index/:mode?', function (e) {
         app.currentView = 'block';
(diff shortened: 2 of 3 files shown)


alicoding:
@k88hudson R?

k88hudson:
Nice!

k88hudson:
actually, let me test one more thing

k88hudson:
This only works for the case of adding a new block; if an existing block is added, and cancel is pressed, it should just cancel the edits, not remove the block. In this case, it should also return to the /make screen, not /add.

alicoding:
@k88hudson based on this ticket though https://github.com/mozilla/webmaker-app/issues/956 they are suggesting we are going back to /add since the user might not finish adding or still exploring each block that they are trying to add.

alicoding:
@k88hudson I might have to go back and look at it again, but at the time when I worked on this bug I felt like the block was added immediately and as they modify the block then we update it in the object.

k88hudson:
@alicoding going back to /add makes sense after adding and then cancelling a block, but not after editing an existing block and then pressing cancel. So to be clear, your implementation is 100% correct for the "new block" case, but not for the edit an old block case. Make sense?

alicoding:
@k88hudson Oh! I didn't test on modifying the block! Oops! Now I have to make sure I cover this part...

alicoding:
@k88hudson what do you think of this?

k88hudson:
@alicoding mind doing a quick rebase?

alicoding:
@k88hudson done :)

thisandagain:
Great. Thanks @alicoding

Issue #1250, opened 2015-02-20
Android setup docs links


pippinlee:
Was going through setting up Cordova/Android SDK and found a few small errors in docs.

All related to this [Android docs](http://webmaker-mobile-guide.mofodev.net/_docs/get_started/android.html) page

* Top Android sdk link broken, needs reformatted

* Wrong ngrok website link, should be [https://ngrok.com/download](https://ngrok.com/download)

* Incomplete sentence describing `adb version` command

Sorry if I missed these discussed elsewhere. Happy to submit a PR if these should be fixed.

thisandagain:
Great catch @pippinlee. Thanks! Happy to accept a PR if you are up for it!
:smile:

k88hudson:
:scroll:  :scroll: :dizzy: 

 (^ docs emoji)

thisandagain:
Resolved by https://github.com/mozilla/webmaker-app/pull/1256

PR #1256, opened 2015-02-20
Clean up Android setup docs #1250

pippinlee:
Clean up info for using ngrok, android developer bridge, and SDK for Android setup docs

📜:dizzy:

### DIFF
--- a/_docs/get_started/android.md
+++ b/_docs/get_started/android.md
@@ -64,7 +64,7 @@ Unfortunately, Webview in Android 4.2 does not ship with a debugger – we will 
 npm install -g weinre
 ```
 
-In order to make debugging easier, you will also want to tunnel to the local port that `weinre` runs on (the alternative is making sure your local computer and phone are on the same network and looking up your IP address, which is time-consuming and requires constant re-configuration). I use `ngrok` to do this. You can download it from [the ngrok website](https://ngrok.com/).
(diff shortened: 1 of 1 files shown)


_docs/get_started/android.md
```diff
@@ -64,7 +64,7 @@ Unfortunately, Webview in Android 4.2 does not ship with a debugger – we will
 npm install -g weinre
 ```
 
-In order to make debugging easier, you will also want to tunnel to the local port that `weinre` runs on (the alternative is making sure your local computer and phone are on the same network and looking up your IP address, which is time-consuming and requires constant re-configuration). I use `ngrok` to do this. You can download it from [the ngrok website](`ngrok`).
+In order to make debugging easier, you will also want to tunnel to the local port that `weinre` runs on (the alternative is making sure your local computer and phone are on the same network and looking up your IP address, which is time-consuming and requires constant re-configuration). I use `ngrok` to do this. You can download it from [the ngrok website](https://ngrok.com/).
```
thisandagain: Instead of "I use `ngrok` to do this" perhaps we could say, "You can use `ngrok`  to do this".

thisandagain:
Looks great! One minor adjustment in the line notes.

_docs/get_started/android.md
pippinlee: @thisandagain Wasn't sure if I should touch that phrasing of the copy -- but totally agree. Will update this PR.

thisandagain:
\o/ R+ Thanks @pippinlee !

PR #1262, opened 2015-02-20
bring template.json in line with the updated phone brick

### LINKED ISSUE #1189: TypeError: obj is undefined
*OS:* ALL
*Browser:* ALL
*Screenshot:* N/A

**Repro Steps:**
There are lots of ways to get here, I'll try to offer up a couple
1. Load webmaker app's home (make view)
2. Tap the "Share Community News" template
3. proceed to "start building"

**Expected:** Webmaker app generates a new make and loads the make view for it
**Actual:** `TypeError: obj is undefined` is thrown and the app gets stuck in the loading screen. back button does not change the view. basically have to force quit.

Less urgent STR:
1. Load the app by navigating directly to a make (either by entering in the URL for a make in the URL bar or by refreshing)

**Expected:** Webmaker app loads the make in question and shows the make view for it
**Actual:** `TypeError: obj is undefined` is thrown and the app gets stuck in the loading screen. back button does not change the view. basically have to force quit.

Pomax:
fixes #1189 by making sure the template.json file has the same properties as the updated phone brick does now (`innerHTML`, `number`, and the missing-before-now `color`)

### DIFF
--- a/lib/templates.json
+++ b/lib/templates.json
@@ -651,6 +651,12 @@
                 "label": "Phone #",
                 "type": "string",
                 "value": "+18005555555"
+            },
+            "color": {
+                "label": "Button Color",
+                "type": "color",
+                "value": "#64A8EE",
+                "skipAutoRender": true
             }
         },
         "type": "phone"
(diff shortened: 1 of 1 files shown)


alicoding:
Tested and working great!! I can add any make and revisit without seeing any obj error anymore <3

PR #1265, opened 2015-02-20
extend i18n.get with macro replacement capabilities

### LINKED ISSUE #1255: App share message says {{app.name}} instead of actual app name
Just made an app. Tried to share it via text message. Message says `{{app.name}}` instead of actual app name.

screenshot_2015-02-20-12-20-57

Pomax:
fixes #1255 by adding macro capabilities to i18n.get

### DIFF
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -70,7 +70,7 @@ function Localize () {
 
     };
 
-    self.get = function (key) {
+    self.get = function (key, macros) {
         var dict = self.dictionary[key];
         var defaultLang = self.dictionaries[self.defaultLang][key];
         if (self.dictionaries[model.data.session.locale] !== undefined) {
--- a/lib/i18n.js
+++ b/lib/i18n.js
@@ -79,7 +79,15 @@ function Localize () {
                 return localized;
             }
         }
-        return dict || defaultLang || key;
+        var value = dict || defaultLang || key;
+        if (!!value) {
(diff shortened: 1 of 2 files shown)


lib/i18n.js
```diff
@@ -79,7 +79,15 @@ function Localize () {
...
             }
         }
-        return dict || defaultLang || key;
+        var value = dict || defaultLang || key;
+        if (!!value) {
+            var macroReplace = function (pattern, term) {
+                var replacement = !!macros ? macros[term] : false;
+                return replacement ? replacement : pattern;
+            };
+            value = value.replace(/{{\s*([^}\s]+)\s*}}/g, macroReplace);
```
secretrobotron: You sure you want to leave out whitespace from the group? e.g. `app name` won't work, but `appname` or `app.name` will.

secretrobotron:
Just spoke with @k88hudson a bit about this. The problematic message was altered to remove its macro (https://github.com/mozilla/webmaker-app/blob/master/locale/en_US/webmaker-app.json#L298), and the others don't use the `i18n` lib programmatically.

Her advice was to close this and avoid putting macros in those strings.

@Pomax make sense?

lib/i18n.js
Pomax: good point

Pomax:
yeah that works for me!

PR #1266, opened 2015-02-20
removing burger; adding new icon menu

### LINKED ISSUE #1179: Fix navigation flow / inconsistencies
The current navigation flow & IA doesn't feel quite right as designed and has been a major sticking point during testing over the past week. We need to devise a short-term fix that can be implemented quickly to help reduce this friction.

gvn:
Resolves #1179 

(Suggest adding &w=0 to diff since there's a lot of indentation fixes...)

### DIFF
--- a/views/block/index.js
+++ b/views/block/index.js
@@ -56,8 +56,11 @@ module.exports = view.extend({
             app.updateBlock(index, {
                 attributes: this.$data.block.attributes
             });
-            this.page(this.$data.back);
             this.$data.saveDisabled = true;
+            this.page(this.$data.back);
+        },
+        onCancel: function (e) {
+            global.history.back();
         }
     },
     created: function () {
--- a/views/make/index.js
+++ b/views/make/index.js
@@ -134,6 +134,10 @@ module.exports = view.extend({
             app.removeApp();
             this.page('/profile');
         },
+        trashClick: function () {
+            this.$dispatch('showShim');
+            this.$dispatch('openModalPrompt', {type: 'delete'});
+        },
         publish: function () {
             var self = this;
             var id = self.$root.$data.params.id;
@@ -199,18 +203,6 @@ module.exports = view.extend({
                 self.fadeOut(element);
             }, 3000);
         }
-
-        self.$on('sideMenuDeleteClick', function (event) {
-            self.$dispatch('openModalPrompt', {type: 'delete'});
-        });
-
-        self.$on('sideMenuShareClick', function (event) {
-            self.publish();
-        });
-
-        self.$on('sideMenuDataClick', function (event) {
-            this.mode = 'data';
-        });
 
         self.$on('onConfirmClick', function (event) {
             console.warn('Deleting app!');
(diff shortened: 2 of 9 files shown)


gvn:
This also depends on https://github.com/k88hudson/ionicons/pull/1 being merged. I used `npm link` locally to get all the new glyphs.

k88hudson:
merged :+1:

k88hudson:
now to test this

views/make/index.js
```diff
@@ -134,6 +141,13 @@ module.exports = view.extend({
             app.removeApp();
             this.page('/profile');
         },
+        trashClick: function () {
+            this.$dispatch('showShim');
+            this.$dispatch('openModalPrompt', {type: 'delete'});
+        },
+        shareClick: function () {
```
k88hudson: why not use publish directly?

views/make/index.js
```diff
@@ -73,6 +73,13 @@ module.exports = view.extend({
         goPreviousView: function (e) {
             global.history.back();
         },
+        onBackClick: function () {
+            if (this.mode !== 'edit') {
+                this.goPreviousView();
+            } else {
+                this.page('/profile');
```
k88hudson: i think actually this should just go back in history, not to profile anymore

gvn: That's actually a whole can of worms because "back in history" can mean back to sub views of the editor (like Add A Brick), which means you can never get out.

gvn: I'm def open to ideas here, but this is basically the functionality from the previous slideout nav.

views/make/index.js
gvn: Yeah, good call.

views/make/index.js
k88hudson: Ah! good point. Can you just change this to the app admin details view instead? That would make more sense than the profile

gvn:
I refactored to use `global.history.back()` in all cases which seems to work a lot better.

k88hudson:
@gvn awesome, will take a look asap

k88hudson:
@gvn the `global.history.back()` is problematic if the previous screen was the editor or add panel :(

gvn:
How so? Also did you start from the root of the app?

k88hudson:
STR: open app editor > tap on block, go to block editor > save, go back to app editor  > press back > 

expected: should go back to app admin
actual: goes back to block editor

gvn:
Ok, that should be fixed now!

gvn:
Ugh...one more issue found. Hang on...

gvn:
Should I merge @k88hudson ?

k88hudson:
@gvn lemme give it one more go first, doing that now

views/block/index.js
```diff
@@ -56,8 +56,11 @@ module.exports = view.extend({
             app.updateBlock(index, {
                 attributes: this.$data.block.attributes
             });
-            this.page(this.$data.back);
             this.$data.saveDisabled = true;
+            global.history.back();
```
k88hudson: I think this should for sure go to the /make page

k88hudson:
R+! Nice work. :cake:

PR #1275, opened 2015-02-21
Remove old locale files

k88hudson:
This removes old locale files before downloading new ones, so we don't accidentally use old files that no longer exist.


thisandagain:
R+ :shipit:

PR #1286, opened 2015-02-23
Tapping on Featured/Newest shouldn't be a toggle

### LINKED ISSUE #1100: Tapping on Featured/Newest shouldn't be a toggle
STR:

 - tap on discover icon to go to discover sheet
 - tap on newest to switch to newest (good)
 - tap on newest toggles back to featured (bad)

Tapping on featured should move to featured *or stay on featured if that's the current tab*. Same for newest.

gesa:
Fixes #1100

### DIFF
--- a/components/switch/index.js
+++ b/components/switch/index.js
@@ -14,7 +14,13 @@ module.exports = {
     },
     methods: {
         onClick: function (event) {
-            this.value = !this.value;
+            if (this.options[(0 + this.value)] === event.target.textContent) {
+                // is the current option the same as the option that was just
+                // tapped on? if so, do nothing.
+                return false;
+            } else {
+                this.value = !this.value;
+            }
         }
     }
 };


components/switch/index.js
```diff
@@ -14,6 +14,9 @@ module.exports = {
     },
     methods: {
         onClick: function (event) {
+            if (this.options[(0 + this.value)] === event.target.textContent) {
```
gvn: Could probably use a comment

components/switch/index.js
```diff
@@ -14,6 +14,9 @@ module.exports = {
...
+                return false;
+            }
             this.value = !this.value;
```
gvn: Might want to explicitly make this an else case for more clarity...

gvn:
A couple nits/suggestions, but R+!

Issue #1304, opened 2015-02-24
Back button doesn't work on "start from scratch" app.


secretrobotron:
It attempts to go back, but there is no previous page, so you return to the page you're already on.

In fact, every time you push the back button, a new app is stored in your "My Apps" section.

alicoding:
@k88hudson if you did start I can totally close my PR. Sorry about that.

PR #1320, opened 2015-02-24
[#1304] Fix back button bug on new blank template

k88hudson:


k88hudson:
This error was happing because landing on the `/template-preview` page when `id` is `blank` results in an auto-create, but the history still contains `/template-preview` in the path.

lib/view.js
```diff
@@ -25,6 +25,17 @@ module.exports = Vue.extend({
...
             e.preventDefault();
             return false;
+        },
+        createAppFromTemplate: function (templateId) {
+            var self = this;
+            self.$root.isReady = false;
+            var app = self.$root.storage.createApp({
+                template: templateId
+            });
+            setTimeout(function () {
```
gvn: Can you use a setInterval to check whatever you're trying to ensure is ready? This looks like a potential race condition...

k88hudson: This is just a visual thing, so the app being created feels more like a deliberate action. It's synchronous because if we make it async/in a callback, it won't work in an offline state

gvn:
Patch worked for me. Recommend changing `setTimetout` to `setInterval` if possible, but otherwise R+! :hamburger:

PR #1328, opened 2015-02-24
Fix #1310 - Add aliases support for l10n

### LINKED ISSUE #1310: Alias locales for which Android doesn't have specifics
If a user's locale is bn-bd or bn-in, google reduces both to just "bn" on Android. Webmaker app doesn't choose the correct locale because we have "bn-bd".

We need a alias for bn-XX -> bn. Or, simply bn-bd to bn for now.

@thisandagain ok to route bn-in to bn as well?

alicoding:
Closes #1310

### DIFF
--- a/views/healthcheck/index.html
+++ b/views/healthcheck/index.html
@@ -12,11 +12,11 @@
   <dt>CURRENT_LANG</dt>
   <dd>
     <select class="dropdownChoiceSelect" v-model="currentLang" v-on="change: selectedLang">
+      <option value="use-default">Use device default</option>
       <option v-repeat="options" value="{{$value}}" class="option-{{$index}}">{{ langmap($value) }}</option>
     </select>
   </dd>
 </dl>
-
 <dl>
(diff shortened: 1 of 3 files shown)


alicoding:
@k88hudson R? [[flagged!]](http://plzkthx.herokuapp.com)

views/healthcheck/index.html
```diff
@@ -15,8 +15,10 @@
       <option v-repeat="options" value="{{$value}}" class="option-{{$index}}">{{ langmap($value) }}</option>
```
k88hudson: Instead of adding another button, just add another element above here like this
```
<option value="use-default">Use device default</option>
```

and then in the `selectedLang` function, check if the `e.target.value === 'use-default'`

alicoding:
Done :)

secretrobotron:
@k88hudson @alicoding good to merge?

alicoding:
@secretrobotron I guess just need one last round of R? from @k88hudson and it should be good to merge.

PR #1338, opened 2015-02-24
Fix issue #1305 - Can't edit block on start from scratch app

### LINKED ISSUE #1305: Can't edit block on "start from scratch" app
STR
1. "Start from Scratch"
2. Add any block
3. Click on gear icon
4. Change the icon colour
5. Click the save button
6. The "edit" icon for the block you created is now covered by the toucan's message :bird:

alicoding:
Closes #1305

### DIFF
--- a/views/make/index.js
+++ b/views/make/index.js
@@ -82,6 +82,7 @@ module.exports = view.extend({
         onSave: function (e) {
             this.goBack(e, false);
             this.updateName(this.$data.app.name);
+            this.hideSpeech();
         },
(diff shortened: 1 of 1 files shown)


Pomax:
hm, I don't see a difference on the start from scratch template with this fix. In both cases, after adding the first block, it takes about a second for the textbox to hide itself, without user interaction.

Pomax:
rescinded. was able to reproduce and this PR fixes it.

PR #1342, opened 2015-02-24
Fix #1312 - Enable Hindi locale

### LINKED ISSUE #1312: Hindi not in the list under Developer Health Check > CURRENT_LANG
I downloaded and installed the app from http://mzl.la/installwm But Hindi is not listed under Developer Health Check > CURRENT_LANG. 

I even changed my phone's locale to Hindi, but still could not see Hindi listed in the CURRENT_LANG.
My phone is running Android 4.4.4 and a Cynogenmod Kernel.

alicoding:
Closes #1312

### DIFF
--- a/gulp/download-locales.js
+++ b/gulp/download-locales.js
@@ -2,7 +2,7 @@ var downloadLocales = require('webmaker-download-locales');
 
 // Which languages should we pull down?
 // Don't include en-US because that's committed into git already
-var supportedLanguages = ['id', 'en_CA', 'es_CL', 'fr', 'nl', 'es_MX', 'cs', 'sv', 'bn_BD', 'sw'];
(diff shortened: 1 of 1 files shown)


alicoding:
@k88hudson R?

thisandagain:
I'm not sure if it's related to this patch, but while testing I found a fairly serious issue.

STR:
- Launch app
- Open developer menu
- Change language to Hindi
- Close developer menu
- Try to change tabs

Result:
```
ReferenceError: नाम is not defined
index.js:16104:0
```

thisandagain:
I can confirm that this does not happen with `french`.

alicoding:
It is an error on Translation directly, so this is now fixed on Transifex and it should be updated once we do another pull again (in an hour).

thisandagain:
Thanks @alicoding !

PR #1345, opened 2015-02-24
Fix #1344 - Default value for link block

### LINKED ISSUE #1344: make the link element either have no placeholders, or prefilled defaults
adding a link element, and editing it, suggests it has default values "mozilla.org" and "mozilla". As a demo user, that's good enough for me and I hit save. Instead of doing what I expect (make a link "mozilla" to "mozilla.org"), this now gives me what looks like a broken element.

We should probably turn those placeholders into actual values, so that if a user hits "edit", we copy the values over from `placeholder` to `value`, and now if the user saves, they have a working link element.

(note that we do this for other elements already, like the phone element)

alicoding:
Closes #1344

### DIFF
--- a/blocks/link/index.js
+++ b/blocks/link/index.js
@@ -8,14 +8,12 @@ module.exports = {
             link: {
                 label: 'URL',
                 type: 'input',
-                value: '',
-                placeholder: 'https://mozilla.org'
+                value: 'https://mozilla.org'
             },
             text: {
               label: 'Link Text',
               type: 'input',
-              value: '',
-              placeholder: 'Mozilla'
+              value: 'Mozilla'
             }
         }
     }


alicoding:
@Pomax R?

blocks/link/index.js
```diff
@@ -8,13 +8,13 @@ module.exports = {
             link: {
                 label: 'URL',
                 type: 'input',
-                value: '',
+                value: 'https://mozilla.org',
                 placeholder: 'https://mozilla.org'
```
Pomax: value overwrites placeholder, so these can go with this change.

blocks/link/index.js
```diff
@@ -8,13 +8,13 @@ module.exports = {
...
             },
             text: {
               label: 'Link Text',
               type: 'input',
-              value: '',
+              value: 'Mozilla',
               placeholder: 'Mozilla'
```
Pomax: overwritten by value

alicoding:
r+ by @Pomax on IRC

PR #1352, opened 2015-02-25
fix back behavior

### LINKED ISSUE #1331: Back button on /make/:id has unexpected behaviour
Right now if you build a template, edit a component, change its values, save those changes, and then hit "back", you're taking back to the "change the values" screen, rather than to the "build the template" screen. It feels like we're doing this:

    build template -> history.pushState
    tap component edit button
    get into edit view -> history.pushState
    save change
    get back to app view -> history.pushState
    tap component edit button
    get into edit view -> history.pushState
    save change
    get back to app view -> history.pushState

when we should probably be doing:

    build template -> history.pushState
    tap component edit button
    get into edit view -> history.pushState
    save change
    get back to app view -> remove edit state from history
    tap component edit button
    get into edit view -> history.pushState
    save change
    get back to app view -> remove edit state from history

That way, in order to "get back" to things like the template menu, you're not first tapping "back" through however many edits you made to your app.

(found during https://github.com/mozilla/webmaker.org/issues/1261 QA checking)

gvn:
Resolves #1331

### DIFF
--- a/views/block/index.js
+++ b/views/block/index.js
@@ -57,7 +57,8 @@ module.exports = view.extend({
                 attributes: this.$data.block.attributes
             });
             this.$data.saveDisabled = true;
-            this.page(this.$data.back);
+            global.history.back();
+            global.history.replaceState({}, '', this.$data.back);
         },
         onCancel: function (e) {
             global.history.back();


Pomax:
small comment about branches that are purely alphanumeric: they can conflict with commit numbers, so a "git checkout gvn/1331" can actually check out a commit that starts with 1331. Often, not an issue. Sometimes, very weird results.

Pomax:
I'm getting `TypeError: str is undefined` errors on if I edit several bricks in succession

gvn:
Hmm. I can't repro that. Here's a screencast, which I'm pretty sure follows your STR:

https://dl.dropboxusercontent.com/u/268451/moz/screencasts/1331.mov

gvn:
What browser/OS were you using?

Pomax:
r+, this was the weirdest git issue I've ever had

gvn:
I swear I'll try to prefix my branch names now...

PR #1370, opened 2015-02-26
Fix title bar centering

k88hudson:


thisandagain:
Looks great. The only problem where I'm still seeing an issue is the "Add Content" screen in which the title is off to the left and the "soft" back button does not work.

thisandagain:
My bad ... don't know how to `git` apparently. :stuck_out_tongue_winking_eye:

PR #1373, opened 2015-02-26
Fix #1367 - Edit app page is empty

### LINKED ISSUE #1367: Edit app page is empty
STR:

1. Open the app
2. Go to "Make" tab
3. Choose "Start from Scratch"
4. Click the gear button

Expected: 
To see the edit title and the logo maker

Actual result:

Blank

screen shot 2015-02-26 at 11 40 19

alicoding:
Closes #1367

### DIFF
--- a/views/make/index.html
+++ b/views/make/index.html
@@ -7,7 +7,7 @@
         <img src="images/toucan1.svg"/>
     </div>
 
-    <main v-show="noBlock" v-class="edit-container: mode === 'edit', settings-container: mode === 'settings'">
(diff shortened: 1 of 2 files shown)


alicoding:
This was caused by https://github.com/mozilla/webmaker-app/commit/d9cd189d56822ff0c07fc08642d4ce6fafea2368 which we were trying to remove the scroll bar. I'm removing the padding when there is no block which is what we intended to make.

alicoding:
@thisandagain R?

thisandagain:
The settings view now loads, but the margins are incorrect / do not match settings everywhere else.

alicoding:
Good catch @thisandagain !!!

thisandagain:
R+ :floppy_disk:

PR #1390, opened 2015-02-26
Fixes the build.

gesa:


alicoding:
Good to merge after Travis give green light!

Tested on my Android, Chrome and Firefox and it is working as expected.

Issue #1382, opened 2015-02-26
Update Android docs with Cordova link and Flash JDK note


pippinlee:
Small fixes to [Developing on Android](http://webmaker-mobile-guide.mofodev.net/_docs/get_started/android.html) docs:

* The "clone mozilla/webmaker-app-cordova" instruction should link to repo like the webmaker-app instruction above it does. Should link to this repo:
[github.com/mozilla/webmaker-app-cordova](https://github.com/mozilla/webmaker-app-cordova)

* Installing the basic Android SDK on a new machine I noticed the `android` command won't work unless you have the Java Development Kit (JDK) installed. Also requires Java to be installed, but most people likely have Java installed already. Is it worth adding a small note in the docs for the JDK?

Happy to send a PR for these if y'all agree.

PR #1396, opened 2015-02-27
Android docs #1382

pippinlee:
* Add Cordova repo link

* Clarify JDK dependency 

Without the Java Development Kit installed users won't be able to run the android sdk manager.


k88hudson:
:+1:

PR #1397, opened 2015-02-27
Enable other locales that fully translated on Transifex

alicoding:


alicoding:
@thisandagain or @k88hudson R?

thisandagain:
R+

Issue #1302, opened 2015-02-24
Data tests sometimes timeout on travis


k88hudson:
Might be worth looking into why, or at least increasing the timeout
```
 Data save should return no errors:
     Error: timeout of 2000ms exceeded
      at null.<anonymous> (/home/travis/build/mozilla/webmaker-app/node_modules/gulp-mocha/node_modules/mocha/lib/runnable.js:158:19)
      at Timer.listOnTimeout [as ontimeout] (timers.js:112:15)
```

PR #1411, opened 2015-03-01
Fix #1409 - Update WebIDE docs

### LINKED ISSUE #1409: Web IDE docs - no longer requires FF Dev browser, update debug
[WebIDE docs](http://webmaker-mobile-guide.mofodev.net/_docs/get_started/web_ide.html) can be updated to reflect WebIDE browser support.

* Currently docs say `"In order to use Firefox WebIDE, you need to have Firefox Developer Edition installed."`
    * Now the WebIDE can be run from Firefox 36.0+ as per: [https://developer.mozilla.org/en-US/docs/Tools/WebIDE](https://developer.mozilla.org/en-US/docs/Tools/WebIDE)

* The WebIDE editor also now uses the debug icon instead of pause for debugging the app.

Worth adding?

pippinlee:


pippinlee:
FYI to whoever reviews this, Travis failing is due to timeout issue. Related to #1302

k88hudson:
@pippinlee thanks for doing this. just to be on the safe side we're going to hold off on merging any PRs until after MWC, but we will review/merge as soon as MWC is over :+1:

PR #1418, opened 2015-03-02
"create a survey" template opens an app called "my teaching app"

### LINKED ISSUE #1407: "create a survey" template opens an app called "my teaching app"

gesa:
Fixes #1407


k88hudson:
Good catch!

PR #1433, opened 2015-03-17
refactor the App object to a normal object instead of inline-defined object

### LINKED ISSUE #1425: lib/storage defines an object type App inline, instead of as a regular object type
https://github.com/mozilla/webmaker-app/blob/master/lib/storage.js#L60 has a full object type with prototype inline-defined, which means it gets redefined, locally scoped, on each call. We should *really* rewrite that to be a lib-scoped single object `App`, with proper construction.

As is, JSCS will trip over the nested this/that/self bindings, and the only real way to fix that is by lifting this code to the correct depth (most of that can be fixed because almost all the `self` references inside App aren't actually necessary)

Pomax:
fixes #1425 by moving the `App` object definition into its own require-able file, rather than inlining the definition to be redeclared when `Storage` functions are called.

This passes lint/jscs/beautify + unit tests, but it also seems like the kind of refactor that is going to reveal bits that we didn't have test coverage for, so this will need some `RealHuman™` testing before we land it.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -1,0 +1,112 @@
+var clone = require('clone');
+
+function App(appref, id, fbRef, data, _firebase, _newBlock) {
+  this.appref = appref;
+  this.id = id;
+  this.data = data || null;
+  this.firebase = fbRef || _firebase.child(id);
+  this.firebase.on('value', this.broadcast, this);
+  this.newBlock = _newBlock;
+}
+
+App.prototype = {
+  broadcast: function (snapshot) {
+    this.data = this.appref.processSnapshot(snapshot);
+    this.appref._vm.$broadcast(snapshot.key(), this.data);
+  },
+
+  update: function (properties) {
+    properties = JSON.parse(JSON.stringify(properties));
+    this.firebase.update(properties, this.appref._onSync);
+  },
+
+  updateBlock: function (id, properties) {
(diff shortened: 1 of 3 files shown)


lib/app.js
```diff
@@ -0,0 +1,112 @@
+var clone = require('clone');
+
+function App(appref, id, fbRef, data, _firebase, _newBlock) {
```
Pomax: this is one part I'm on the fence about. Constructor args? We never use more than this. Options object? Cleaner code... marginally.

lib/storage.js
```diff
@@ -46,9 +47,7 @@ Store.prototype._createAppRef = function (id, fbRef, data) {
     this._removeAppRef(id);
   }
 
-  var process = this._processSnapshot;
-
-  function newBlock(blockId) {
+  var app = new App(this, id, fbRef, data, _firebase, function newBlock(blockId) {
```
Pomax: inlining this function is certainly ugly, but if I leave it as "defined in this function" as it is right now, jshint (rightly) compains it's not actually being used in the file that defined it.

Pomax:
Flagging @thisandagain and @gvn for some testing

thisandagain:
I propose we close this as per the work going into v1 right now. Sound ok to you @Pomax ?

PR #1437, opened 2015-03-24
Add show in discovery in settings panel

### LINKED ISSUE #1267: "Show in discovery" should be in settings panel
Following from #1179 (where there's a settings panel!), the "show in discovery gallery" checkbox should be in the settings panel. Currently, it's hard to find.

ryandang:
fixes #1267


thisandagain:
R+ Thanks @ryandang ! :smile:

PR #1440, opened 2015-03-26
Add delete button in preview app for admin

### LINKED ISSUE #1436: Have a delete button beside open my app?
I know you can delete the app after you open it but I think it's would be nice if we have a delete button before open the app.

we'll have the delete button in red color
delete

ryandang:
fixes #1436


thisandagain:
R+ :cow2:

PR #1484, opened 2015-04-07
Clean up & Organize Router

ryanw-ny:
Everything is nice and tidy, and in separate packages, to avoid confusion while working with it.


ryanw-ny:
@thisandagain This should be good to go now!

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
```diff
@@ -2,29 +2,27 @@
...
-        return singleton;
-    }
 
     @Override
     public void onCreate() {
         super.onCreate();
         singleton = this;
 
-        Log.v("[Webmaker]", "Application created.");
+        Log.v("[WebMaker]", "Application created.");
```
thisandagain: "Webmaker" the brand is never PascalCase.

app/src/main/java/mozilla/org/webmaker/router/Router.java
```diff
@@ -160,24 +79,21 @@ public void map(String format, RouterCallback callback) {
     /**
      * Map a URL to open an {@link Activity}
      * @param format The URL being mapped; for example, "users/:id" or "groups/:id/topics/:topic_id"
-     * @param klass The {@link Activity} class to be opened with the URL
+     * @param clazz The {@link Activity} class to be opened with the URL
      */
-    public void map(String format, Class<? extends Activity> klass) {
-        this.map(format, klass, null);
+    public void map(String format, Class<? extends Activity> clazz) {
+        this.map(format, clazz, null);
```
thisandagain: LOL. :+1:

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
ryanw-ny: @thisandagain Not sure how I missed that.. will fix!

thisandagain:
Cool. So I think the only question that this raises is documenting how devs can trigger `router.open` outside of the `WebmakerApplication` class. We'll need this in order to build up the interface to JS this week. Mind adding some minimal docs to the top of the Router?

PR #1485, opened 2015-04-07
merging in protozoom

gvn:


k88hudson:
@gvn looks like you need to add react-draggable to package.json

gvn:
Oops...must have lost that during my package.json tweaking.

package.json
```diff
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "react": "^0.13.1",
+    "react-draggable": "https://github.com/gvn/react-draggable#master",
```
k88hudson: you should use `git://` as per npm docs

k88hudson: i.e. `git://github.com/gvn/react-draggable.git#master`

www_src/pages/section-2/section-2.jsx
```diff
@@ -1,14 +1,210 @@
...
+
+// TEMP: Just used for generating random starter layouts
+function generateLayout (width, height) {
+  var layout = [];
+
+  for (var y = 0; y < height; y++) {
+    layout.push([]);
+
+    for (var x = 0; x < width; x++) {
+      layout[y].push(Math.random() > 0.25 ? null : '/img/page-screenshot.png');
```
k88hudson: don't use absolute paths, since they won't work for android (everything is inside a www/assets folder)

www_src/pages/section-2/section-2.jsx
```diff
@@ -1,14 +1,210 @@
...
+          nodes.push(
+            <Slot x={x} y={y} perRow={this.state.layout[0].length}/>
+          );
+        }
+      }
+    }
+
+    var gridTransform = {
+      transform: 'scale(' + this.state.zoom + ')',
+      WebkitTransform: 'scale(' + this.state.zoom + ')'
```
k88hudson: isn't this supposed to be lower case?

www_src/pages/section-2/section-2.jsx
```diff
@@ -1,14 +1,210 @@
...
+    }
+
+    return (
+      <div className="grid" style={gridTransform}>
+        {nodes}
```
k88hudson: when you render lists of nodes you need to add a `key` property to each one or else React will display a warning. you can just use the index in the array

k88hudson:
Great start, I left some minor comments. In the android build, your zooming via +/- buttons works fine but as soon as you tap the inner area it goes blank. No console errors. Feel free to address in a follow up ticket, or if you can't reproduce in Chrome on device, when you get set up on Android studio

image

www_src/pages/section-2/section-2.jsx
gvn: No, that's how they name vendor prefixes in React for some reason...

PR #1489, opened 2015-04-08
[Fix #1488] Destroy webview

### LINKED ISSUE #1488: Web views are not destroyed after detaching the view
Instead, they detach and stick around

k88hudson:
I think this should do the trick


thisandagain:
R+ Thanks.

PR #1504, opened 2015-04-14
+ zoom in on tile, + add ability to zoom back out

gvn:


www_src/pages/section-2/section-2.jsx
```diff
@@ -62,16 +66,54 @@ var Slot = React.createClass({
...
+      zoom: this.state.containerWidth / this.slotWidth,
+      cameraX: cameraX,
+      cameraY: cameraY
     });
   },
   getInitialState: function () {
     var layout = generateGrid(3, 3);
 
     return {
-      zoom: this.props.initialZoom,
```
Pomax: you're still referencing `this.state.zoom` in places, so either those need fixing, or `zoom` needs to stay in the initial state

www_src/pages/section-2/section-2.jsx
```diff
@@ -259,35 +304,6 @@ var Grid = React.createClass({
...
-    return (
-      <div className="segmented-control">
-        <button onClick={ this.increase }><span>+</span></button>
-        <button onClick={ this.decrease }><span>-</span></button>
-      </div>
-    );
-  }
-});
-
 var App = React.createClass({
```
Pomax: sortof-kindof nit: can this be its own file?

Pomax:
This is doing crazy things in Firefox, with tiles completely misaligning. Technically transforms should work cross browser, so it feels like this is worth making sure works in FF and Chrome instead of just Chrome.

gvn:
Ah...Firefox handles subpixels differently (actually every browser handles them differently). 

Should be fixed now...

www_src/pages/section-2/section-2.jsx
gvn: Planning to break all these classes out once things are a little more solid...

www_src/pages/section-2/section-2.jsx
```diff
@@ -28,6 +31,7 @@ var Page = React.createClass({
 
     return (
       <div
+        onClick={this.handleClick}
```
Pomax: does it make sense to just say `onClick={this.props.onClick}` instead?

www_src/pages/section-2/section-2.jsx
```diff
@@ -170,19 +212,19 @@ var Grid = React.createClass({
     this.gridWidth = undefined;
     this.gridHeight = undefined;
 
+    this.slotWidth = undefined;
```
Pomax: we should probably move these to the `componentWillMount` function instead, instead of redeclaring them on render() every time.

www_src/pages/section-2/section-2.jsx
```diff
@@ -170,19 +212,19 @@ var Grid = React.createClass({
...
 
     // Parse aspect ratio
     var widthAR = parseInt(this.props.aspectRatio.split(':')[0], 10);
     var heightAR = parseInt(this.props.aspectRatio.split(':')[1], 10);
 
-    var slotWidth;
-    var slotHeight;
-
     // Determine if a slot has a neighboring page in any cardinal direction
     function hasNeighbors(x, y) {
```
Pomax: should this function be redeclared each time `render` runs? It might make sense to make this a `this.hasNeighbours`

www_src/pages/section-2/section-2.jsx
```diff
@@ -295,6 +314,13 @@ var App = React.createClass({
   changeZoom: function (event) {
     this.refs.masterGrid.zoom(event.amount);
   },
+  showOverview: function () {
+    this.refs.masterGrid.setState({
```
Pomax: it's better to make this a real function instead of calling setState on "not yourself", so calling `this.refs.masterGrid.setValues(1, 0, 0)` and then in Grid having a `this.setValues: function(zoom, cameraX, cameraY) { this.setState(...); }`

Pomax:
UX wise that totally fixed things between Chrome and Firefox! Few more things, mostly related to code organisation.

www_src/pages/section-2/section-2.jsx
gvn: Good call.

www_src/pages/section-2/section-2.jsx
gvn: Is it redeclaring them or just re-assigning them? Is there a perf benefit to moving them into `componentWillMount`?

I'm kind of ambivalent on choosing one placement or the other since they both have roughly equal advantages in my mind.

www_src/pages/section-2/section-2.jsx
gvn: I think it's fine either way, but I moved it into a utility method...

www_src/pages/section-2/section-2.jsx
gvn: How is that better? I mean...it's more abstracted, but `setState` is a setter, no?

gvn:
ibs

PR #1530, opened 2015-04-20
More progress on map view

gvn:


www_src/pages/section-2/add-page.jsx
```diff
@@ -0,0 +1,15 @@
+var React = require('react');
+
+module.exports = React.createClass({
+  render: function () {
+    return (
+      <button
+        onClick={this.handleClick}
```
Pomax: `onClick={this.props.onClick}`? There doesn't seem to be a need for the pass-through

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
+var React = require('react');
+
+var Slot = require('./slot.jsx');
+var Page = require('./page.jsx');
+var AddPage = require('./add-page.jsx');
```
Pomax: Since this is only added once, does this make sense as a component when it's only a single button with a class?

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+
+module.exports = React.createClass({
+  /**
+   * Zoom in to a page tile
+   * @param  {Number} pageX Page's x coordinate in grid
+   * @param  {Number} pageY Page's y coordinate in grid
+   * @param  {Number} pagesWide Number of pages to fit in width-wise
+   * @param  {boolean} doCentering Determine whether or not to move "camera" to center the target
+   */
+  calculateCameraState: function (pageX, pageY, pagesWide, doCentering) {
```
Pomax: hmm, this feels like a slug of code that belongs in its own plain js library, generating the new data based on the current input

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+  getInitialState: function () {
+    // 20 x 20 seems to be max before display glitches
+    var layout = this.generateGrid(3, 3);
+
+    return {
+      layout: layout
+    }
+  },
+  addPageClick: function (event) {
+    var self = this;
```
Pomax: this should not be necessary in React with ES6

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+
+    var width = newLayout[0].length;
+    var height = newLayout.length;
+
+    // The new page tile's x and y coords may shift if the grid is expanded
+    var newPageX = event.x;
+    var newPageY = event.y;
+
+    // Detect what grid boundaries the given tile touches
+    function edgeDetect(x, y) {
```
Pomax: multiple function declarations in this call suggests this should be a .js lib on its own instead of being intimately tied to the React component. At the very least, those functions can be declared at the same level as addPageClick, so they don't get redefined every time addPageClick is called.

www_src/pages/section-2/grid.jsx
gvn: I'm not using ES6...

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+      return boundaries;
+    }
+
+    // Expand the grid in the specified direction
+    function expandGrid(direction) {
+      function buildEmptyRow(length) {
+        var row = [];
+
+        for (var i = 0; i < length; i++) {
+          row.push(null);
```
Pomax: proper pet-peeve nit: Javascript has no use for null, it already has undefined, but even that is usually unnecessary since JS also has `false`

www_src/pages/section-2/grid.jsx
gvn: Hrmmm...this is very tied to the Grid, so I don't agree.

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+      if (y === newLayout.length - 1) {
+        boundaries.push('bottom');
+      }
+
+      return boundaries;
+    }
+
+    // Expand the grid in the specified direction
+    function expandGrid(direction) {
+      function buildEmptyRow(length) {
```
Pomax: this shouldn't have to be an inline function that is redeclared each time expandGrid is called.

www_src/pages/section-2/grid.jsx
gvn: Yeah, good call, probably doesn't need to be a class as is.

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+    if (edgeDetect(event.x, event.y).length) {
+      edgeDetect(event.x, event.y).forEach(function (direction, index, array) {
+        expandGrid(direction);
+      });
+    }
+
+    this.setState({
+      layout: newLayout
+    }, function () {
+      self.calculateCameraState(newPageX, newPageY, self.displayState.pagesWide, true);
```
Pomax: you don't need `self` here, React already ensures correct context rebinding in `setState` callbacks

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+  },
+  componentDidUpdate: function () {
+    var elGrid = this.getDOMNode();
+
+    // Force fixed width & height on grid
+    elGrid.style.width = this.gridWidth + 'px';
+    elGrid.style.height = this.gridHeight + 'px';
+  },
+  componentDidMount: function () {
+    this.gridWidth = undefined;
```
Pomax: setting all of these to undefined seems less useful than setting them to "defaults" instead (like zeroes across the board).

www_src/pages/section-2/grid.jsx
gvn: Yeah, I'm planning on moving them into utility methods at the very least pretty soon.

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+    var previousTranslateTransform = ' translate3d(' + this.displayState.previousCameraX + 'px, ' + this.displayState.previousCameraY + 'px, 0)';
+
+    var finalTransform = scaleTransform;
+
+    if (typeof this.displayState.cameraX === 'number' && typeof this.displayState.cameraY === 'number') {
+      finalTransform += (' '  + translateTransform)
+    }
+
+    if (animateZoom) {
+      elGrid.classList.add('animated');
```
Pomax: this feels like not taking advantage of React. If this element needs new classes, those should get set in `render` by using something like

```
var cn = require("classnames"); // npm hosted helper lib
render: function() {
  ...
  var elGridClasses = cl({
    elGrid: true,
    animated: animateZoom,
    ...
  });
  return <div classNames={elGridClasses} ...
    ...
}
`

File: www_src/pages/section-2/grid.jsx [hunk 11]
```diff
@@ -0,0 +1,382 @@
...
+        elGrid.style.transform = scaleTransform + ' ' + translateTransform;
+
+        setTimeout(function() {
+          elGrid.classList.remove('animated');
+        }, 300);
+      }, 1);
+    }
+  },
+  render: function () {
+    var self = this;
```
Pomax: should not be needed

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+
+    this.tilesPerRow = layout[0].length;
+    this.tilesPerCol = layout.length;
+
+    // Parse aspect ratio
+    var widthAR = parseInt(this.props.aspectRatio.split(':')[0], 10);
+    var heightAR = parseInt(this.props.aspectRatio.split(':')[1], 10);
+
+    // Try to fit grid in viewport by constraining to the width
+    this.slotWidth = (self.state.containerWidth / this.tilesPerRow);
```
Pomax: no reason for `self` here

www_src/pages/section-2/grid.jsx
gvn: It doesn't have to be, but it makes more sense IMO because it's only used within the context of `expandGrid`.

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+    this.slotHeight = this.slotWidth * (heightAR / widthAR);
+
+    // If the height overflows, then constrain by height instead
+    if (this.slotHeight * this.tilesPerCol > self.state.containerHeight) {
```
Pomax: no need for `self` here

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+    var slotStyle = {
+      width: this.slotWidth + 'px',
+      height: this.slotHeight + 'px'
+    }
+
+    this.gridWidth = this.slotWidth * this.tilesPerRow;
+    this.gridHeight = this.slotHeight * this.tilesPerCol;
+
+    // Render slots
+    for (var y = 0; y < layout.length; y++) {
```
Pomax: these edges were caught as `this.tilesPerCol` and `this.tilesPerRow`, it's nicer to use those here.

www_src/pages/section-2/grid.jsx
```diff
@@ -0,0 +1,382 @@
...
+      for (var x = 0; x < layout[0].length; x++) {
+        if (layout[y][x]) {
+          nodes.push(
+            <Slot x={x} y={y} style={slotStyle} key={ y + '-' + x }>
```
Pomax: all thee Slot generation blocks use mostly the same code, so there's a fair bit of cleanup we can do here, especially because JSX uses ES6 style property dereferencing:
```
var key = [y,x].join('-');
var slotProps = { x: x, y: y, style: slotStyle, key: key };
var slotContent = false;
var clickHandler = this.onPageClick.bind(this, {x:x, y:y});
if (layout[y][x]) {
  slotContent = <Page onClick={ clickHandler } screenshot={layout[y][x]} />
} else if (...) {
  slotContent = <AddPage onClick={ clickHandler } />
}
nodes.push(<Slot {...slotProps}>{ slotContent }</Slot>);
```

www_src/pages/section-2/page.jsx
```diff
@@ -0,0 +1,19 @@
+var React = require('react');
+
+module.exports = React.createClass({
+  render: function () {
+    var style = {};
+
+    if (this.props.screenshot !== 'EMPTY') {
```
Pomax: this seems like a "magic" string - it's probably better to not rely on a special value but instead simply tap into "was there a screenshot at all passed in". If so, style, if not, no style?

www_src/pages/section-2/grid.jsx
gvn: Good call!

www_src/pages/section-2/page.jsx
gvn: Since this was just a temporary thing anyway I'm going to use an empty object instead, which will eventually become a *meaningful* page object.

Issue #1515, opened 2015-04-15
Implement page front-end


xmatthewx:
Ability to add and transform text element and images (via URL only - a temporary feature).

 - [add element ui](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae0e362d5_2_128)
 - [image selected](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae0e362d5_2_114)
 - [text with bounding box transform](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae0e362d5_2_96)

Pomax:
current "live" version up on http://pomax.github.io/webmaker-app/pages/project/ with PR https://github.com/mozilla/webmaker-app/pull/1538

flukeout:
Some feedback:
* [ ] Often the "shadow" of the element will disconnect, and the gestures only affect the shadow
* [ ] Rotation gestures where both fingers rotate around a point between them produce unexpected results
* [ ] It's possible to pan things off of the tile area, and not grab them again while seeing their shadow
* [ ] When an element is selected, I should be able to deselect it by tapping on empty space in the tile
* [ ] Can we show a 'trash' icon when an element is selected and be able to delete the elemetn?
* [ ] Elements 'jump' a little bit after rotating them, removing second finger, and panning
* [ ] Can we show the 'clipped shadow' only when an element is selected? 
* [ ] We are working under the assumption that a 'swipe' gesture will transition pages, but right now the 'swipe' is being interpreted as an element move if you swipe over an element - we need to resolve this

xmatthewx:
[UI comps](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae0e362d5_2_74)

k88hudson:
I'm going to work on this for a bit if that's OK

xmatthewx:
Page editor interactions need to be optimized. 

**Problem**: User can't deselect an element if their page is piled high with many elements. Add-element button is replaced with pencil. Tap to deselect element would force us to reconsider how we handle z-index. Tapping outside page collides with nav between pages ux-wise.

[...]

PR #1538, opened 2015-04-22
Positionable elements + editor view switching

Pomax:
addresses #1515

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,9 +53,8 @@
     "webpack": "^1.7.3"
   },
   "dependencies": {
-    "classnames": "^1.2.0",
-    "hammerjs": "^2.0.4",
-    "react": "^0.13.1",
+    "classnames": "2.2.0",
+    "react": "0.13.1",
     "react-draggable": "git://github.com/gvn/react-draggable#master",
     "react-hammerjs": "0.2.2",
     "react-router": "0.13.2"
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "webpack": "^1.7.3"
   },
   "dependencies": {
-    "classnames": "2.2.0",
+    "classnames": "1.2.1",
     "react": "0.13.1",
     "react-draggable": "git://github.com/gvn/react-draggable#master",
     "react-hammerjs": "0.2.2",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,9 @@
     "webpack": "^1.7.3"
   },
   "dependencies": {
-    "react": "0.13.1",
+    "classnames": "^1.2.0",
+    "hammerjs": "^2.0.4",
+    "react": "^0.13.1",
     "react-draggable": "git://github.com/gvn/react-draggable#master",
     "react-hammerjs": "0.2.2",
     "react-router": "0.13.2"
--- a/www_src/main.less
+++ b/www_src/main.less
@@ -1,5 +1,6 @@
 @import "normalize";
 @import "project-map";
+@import "positionable";
 
 *,
 *:after,
@@ -94,5 +95,3 @@ body {
     width: 50px;
   }
 }
-
-@import "positionable";
--- a/www_src/main.less
+++ b/www_src/main.less
@@ -94,3 +94,5 @@ body {
     width: 50px;
   }
 }
+
+@import "positionable";
--- a/www_src/pages/project/Positionable.jsx
+++ b/www_src/pages/project/Positionable.jsx
@@ -17,15 +17,6 @@ var Positionable = React.createClass({
     };
   },
 
-  componenWillMount: function() {
-    if(!this.state.interactive) return;
-
-    this.markx = this.state.x;
-    this.marky = this.state.y;
-    this.markangle = this.state.angle;
-    this.markscale = this.state.scale;
-  },
-
   componentDidMount: function() {
     if(!this.state.interactive) return;
     var touchHandler = this.touchhandler = require("./touchhandler")(this);
@@ -123,11 +114,11 @@ var Positionable = React.createClass({
 
   setTransform: function(obj) {
     this.setState({
-      x: obj.x || 0,
-      y: obj.y || 0,
-      angle: obj.angle || 0,
-      scale: obj.scale || 1,
-      zIndex: obj.zIndex || 0
+      x: obj.x,
+      y: obj.y,
+      angle: obj.angle,
+      scale: obj.scale,
+      zIndex: obj.zIndex
     });
   }
 });
--- a/www_src/pages/project/Positionable.jsx
+++ b/www_src/pages/project/Positionable.jsx
@@ -1,0 +1,135 @@
+"use strict";
+
+var React = require("react");
+var classes = require("classnames");
+
+var Positionable = React.createClass({
+
+  getInitialState: function() {
+    return {
+      x: this.props.x || 0,
+      y: this.props.y || 0,
+      scale: this.props.scale || 1,
+      angle: this.props.angle || 0,
+      zIndex: (typeof this.props.zIndex !== "undefined") ? this.props.zIndex : 1,
+      interactive: (typeof this.props.interactive !== "undefined") ? this.props.interactive : true,
+      touchactive: false
+    };
+  },
+
+  componenWillMount: function() {
+    if(!this.state.interactive) return;
+
+    this.markx = this.state.x;
+    this.marky = this.state.y;
+    this.markangle = this.state.angle;
+    this.markscale = this.state.scale;
+  },
+
+  componentDidMount: function() {
+    if(!this.state.interactive) return;
+    var touchHandler = this.touchhandler = require("./touchhandler")(this);
+
+    var dnode = this.getDOMNode();
+    dnode.addEventListener("touchstart", touchHandler.startmark);
+    dnode.addEventListener("mousedown", touchHandler.startmark);
+    dnode.addEventListener("touchmove", touchHandler.panmove);
+    dnode.addEventListener("mousemove", touchHandler.panmove);
+    dnode.addEventListener("touchend", touchHandler.endmark);
+    dnode.addEventListener("mouseup", touchHandler.endmark);
+
+    var onode = this.refs.overlay.getDOMNode();
+    onode.addEventListener("touchstart", touchHandler.secondFinger);
+    onode.addEventListener("touchmove", touchHandler.handleRS);
+    onode.addEventListener("touchend", touchHandler.endSecondFinger);
+  },
+
+  componentDidUpdate: function(prevProps, prevState) {
+    if(!prevState.touchactive && this.state.touchactive) {
+      this.props.onUpdate(this.state);
+    }
+  },
+
+  render: function() {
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -53,7 +53,9 @@
     "webpack": "^1.7.3"
   },
   "dependencies": {
-    "react": "0.13.1",
+    "classnames": "^1.2.0",
+    "hammerjs": "^2.0.4",
+    "react": "^0.13.1",
```
gvn: I usually like to version lock non dev-deps because I don't trust people to not introduce breaking changes. :dancers:

www_src/main.less
```diff
@@ -94,3 +94,5 @@ body {
     width: 50px;
   }
 }
+
+@import "positionable";
```
gvn: Can we keep all our imports at the top?

www_src/pages/project/Positionable.jsx
```diff
@@ -0,0 +1,135 @@
...
+      y: this.props.y || 0,
+      scale: this.props.scale || 1,
+      angle: this.props.angle || 0,
+      zIndex: (typeof this.props.zIndex !== "undefined") ? this.props.zIndex : 1,
+      interactive: (typeof this.props.interactive !== "undefined") ? this.props.interactive : true,
+      touchactive: false
+    };
+  },
+
+  componenWillMount: function() {
```
gvn: Shouldn't this be componen**t**WillMount?

www_src/main.less
Pomax: in this case, no - something clobbers my rules by not namespacing, but I'll happily figure out what as a followup

package.json
Pomax: on the one hand: fair enough. On the other: this is literally why we're using semvers: ^ means "any version that isn't a major version upgrade", which is supposed to guarantee stability.

www_src/pages/project/Positionable.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    var className = classes({
+      positionable: true,
+      touchactive: this.state.touchactive,
+      current: this.props.current
+    });
+
+    return (
+      <div>
+        <div ref="overlay" className="touchOverlay" hidden={!this.state.touchactive} />
```
gvn: Self closing div tag? Assuming React allows this?

www_src/pages/project/Positionable.jsx
```diff
@@ -0,0 +1,135 @@
...
+    });
+  },
+
+  handleRotationAndScale: function(angle, scale) {
+    this.setState({
+      angle: this.transform.angle + angle,
+      scale: this.transform.scale * scale
+    }, function() {
+      if(this.props.onUpdate) {
+        this.props.onUpdate(this.state);
```
gvn: Could these callbacks be fired from `componentDidUpdate` instead?

www_src/pages/project/Positionable.jsx
Pomax: technically, yes, but also technically, I was in the process of deleting this function and... forgot? O_o

www_src/pages/project/Positionable.jsx
```diff
@@ -0,0 +1,135 @@
...
+      x: this.state.x,
+      y: this.state.y,
+      angle: this.state.angle,
+      scale: this.state.scale,
+      zIndex: this.state.zIndex
+    };
+  },
+
+  setTransform: function(obj) {
+    this.setState({
```
gvn: Would it be better to only update properties that are passed in `obj` instead of possibly reverting values to defaults?

www_src/pages/project/Positionable.jsx
Pomax: remember that JSX is **not** HTML. JSX allows self closing syntax on all XML notation for components.

www_src/pages/project/blocks/Link.jsx
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+var React = require("react");
+
+var Link = React.createClass({
+  getInitialState: function() {
+    return {
+      href: this.props.href || "",
+      label: this.props.label || "...",
```
gvn: Label makes me think of the label element. Maybe innerText or something?

www_src/pages/project/Positionable.jsx
Pomax: that would make a lot of sense, thanks.

www_src/pages/project/Positionable.jsx
Pomax: the setTransform is for loading in an element wholesale. It's meant to completely overwrite the element to some earlier saved state.

www_src/main.less
gvn: Yeah, that'd be good to figure out. I just nested my stuff inside of `#map-view` so it should play nice now!

www_src/pages/project/blocks/Link.jsx
Pomax: that seems the wrong reason to change this. It's the correct word for what it's used for, and since `<label>` elements only apply to form inputs, so the context doesn't really allow for that interpretation (and end users never see this value)

www_src/pages/project/Positionable.jsx
gvn: Lolz

www_src/pages/project/Positionable.jsx
gvn: I figured as much...neato.

www_src/pages/project/Positionable.jsx
gvn: :+1:

www_src/main.less
Pomax: let me check if things are still getting clobbered. if not, I'll just change the ordering now.

www_src/pages/project/Positionable.jsx
gvn: K. Personally I'd favor a setter that doesn't clobber existing values if I don't re-specify them every time.

www_src/pages/project/project.jsx
```diff
@@ -2,20 +2,142 @@ var React = require('react');
...
+  getInitialState: function() {
+    return {
+      content: [],
+      currentElement: -1
+    };
+  },
+
   render: function () {
-    return <div className="demo">
+    console.log("rerendering");
```
gvn: Stray console.log...

www_src/pages/project/project.jsx
```diff
@@ -2,20 +2,142 @@ var React = require('react');
...
+    var shadows = this.shadows ? this.shadows : this.formShadows(this.state.content);
+    var positionables = this.formPositionables(this.state.content);
+    return <div className="demo project">
```
gvn: Line break after return would make this more readable...

www_src/pages/project/Positionable.jsx
Pomax: turns out I had a reason for this and forgot - a general purpose update in componentDidUpdate will cause infinite recursions, because this.props.onUpdate causes a rerender in Project, which triggers a componentDidUpdate because of changed props. In this case, the update should only kick in for changed state, so they're triggered specifically as a result of setState

www_src/pages/project/project.jsx
```diff
@@ -2,20 +2,142 @@ var React = require('react');
...
+        "/project/123/1?data=" + JSON.stringify({
+          index: this.state.currentElement,
+          entry: this.state.content[this.state.currentElement]
+        })
+      } href="/pages/editor" hidden={true}>edit</Link>
     </div>
+  },
+
+  formShadows: function(content) {
+    console.log("regenerating shadows");
```
gvn: console

www_src/positionable.less
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  .shadows {
+    position: absolute;
+    height: 0;
+    opacity: 0.2;
+    margin-left: -5px;
+
+    -webkit-filter: grayscale(100%);
```
gvn: We're using autoprefixer so you can just use the standard property names. :dancer:

www_src/positionable.less
```diff
@@ -0,0 +1,88 @@
...
+    width: 100%;
+
+    .touchOverlay {
+      position: fixed;
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      z-index: 2147483647; /* signed 32bit maxint */
+      /* background-color: rgba(0,0,0,0.2); */
```
gvn: Just remove this line?

www_src/positionable.less
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    .positionable {
+      -webkit-user-select: none;
+      -moz-user-select: none;
+      -ms-user-select: none;
+      -o-user-select: none;
+      user-select: none;
```
gvn: See other autoprefixer comment...

www_src/pages/project/Positionable.jsx
Pomax: it's a weird function really, because it shouldn't even be called without all those values specified (but I didn't want to do validation before the schema's settled on)

www_src/positionable.less
```diff
@@ -0,0 +1,88 @@
...
+
+      &.current {
+        margin: -1px;
+        border: 1px dashed black;
+      }
+    }
+  }
+
+  .shadows {
+    overflow: visible!important;
```
gvn: !important makes me a sad panda...

www_src/pages/project/Positionable.jsx
gvn: Wait...so doesn't this run 2 renders in a row then?

www_src/pages/project/Positionable.jsx
gvn: Ok, maybe just leave a TODO comment in there then

www_src/pages/project/blocks/Link.jsx
gvn: Yeah...it just feels weird to me, but carry on.

www_src/pages/project/Positionable.jsx
Pomax: it does, but the second render due to onUpdate won't actually touch the DOM since the updated state already matches what the rerender is asking to have updated, so React's virtual DOM fanciness goes "nothing to change here" and things play nice.

www_src/positionable.less
Pomax: awwwwwwww yeah removing all those dirty vendor prefixes O_O

www_src/positionable.less
Pomax: same here, but not when it's used correctly. This is an override on how the `inner` div behaves, which is far more specifically declared in main.less, which is untouchable in terms of what a single view needs to do. In any other view, that overflow should never be visible, but in this case because of funky element repositioning shadowing we need some important

www_src/pages/project/Positionable.jsx
Pomax: good call, done

gvn:
R+

clear

Issue #1501, opened 2015-04-14
v1 - Package management


ryanw-se:
Right now, it appears that things are just stuffed into the main package and not really organized. There should be a Google spreadsheet, or similar with the planned package layout for the project, to better plan ahead the placement of the classes and make it easy to find certain things.

thisandagain:
Generally resolved in the last re-org of the project.

PR #1542, opened 2015-04-23
Refactor project design, remove duplicate code, implement temporary package structure

ryanw-ny:
This removes any bloated or duplicate code from the application, it also moves things around to add more organization and make it easier to work with in the future. Temporary package structure has been implemented although it will probably be revised down the road as more features get implemented.

This shouldn't impact performance or how the program currently functions, this just cleans things up. All sub-activities have been migrated to their own package and can be easily found in ```mozilla.org.webmaker.activity``` excluding the default activity and web maker activities. 

This closes issue  #1501

/cc @thisandagain

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/MainActivity.java
+++ b/app/src/main/java/mozilla/org/webmaker/MainActivity.java
@@ -29,7 +29,7 @@ public class MainActivity extends Activity implements ActionBar.TabListener {
 
         // Set up the action bar.
         final ActionBar actionBar = getActionBar();
-        assert actionBar != null;
+        if (actionBar == null) throw new NullPointerException("ActionBar has returned null!");
         actionBar.setNavigationMode(ActionBar.NAVIGATION_MODE_TABS);
 
         // Create the adapter that will return a fragment for each of the three primary sections of the activity.
--- a/app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java
+++ b/app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java
@@ -34,7 +34,7 @@ public class WebmakerActivity extends Activity {
     @Override
     public void onDestroy() {
         Log.v("wm", "onDestroy");
-        assert view != null;
+        if (view == null) return;
         view.destroy();
         view = null;
         super.onDestroy();
--- a/app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java
+++ b/app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java
@@ -7,11 +7,11 @@ import android.view.Menu;
 import android.view.MenuItem;
 import android.widget.RelativeLayout;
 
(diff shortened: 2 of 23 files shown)


app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java
```diff
@@ -34,10 +34,9 @@ protected void onCreate(Bundle savedInstanceState) {
     @Override
     public void onDestroy() {
         Log.v("wm", "onDestroy");
-        if (view != null) {
-            view.destroy();
-            view = null;
-        }
+        assert view != null;
```
thisandagain: This assert is removed during runtime which makes this brittle, no?

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
```diff
@@ -3,12 +3,16 @@
...
+import mozilla.org.webmaker.activity.Editor;
+import mozilla.org.webmaker.activity.Map;
+import mozilla.org.webmaker.activity.Project;
+import mozilla.org.webmaker.activity.Tinker;
 import mozilla.org.webmaker.router.Router;
 
 public class WebmakerApplication extends Application {
 
     // Singleton
-    private static WebmakerApplication singleton;
```
thisandagain: Good catch. :+1:

app/src/main/java/mozilla/org/webmaker/fragment/WebviewFragment.java
```diff
@@ -68,34 +44,9 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container, Bundle sa
 
     @Override
     public void onDestroyView() {
-        if (mWebView != null) {
-            mWebView.destroy();
-            mWebView = null;
-        }
+        assert mWebView != null;
```
thisandagain: Same question here.

thisandagain:
Looks great. Just one small question / concern regarding the use of an assertion within the `onDestroyView` lifecycle event.

ryanw-ny:
@thisandagain Good catch with the assertions! I went ahead and replaced all of them with regular conditional statements like suggested and added a check to see if the ActionBar is null and if so throw a error for debugging purposes.

app/src/main/java/mozilla/org/webmaker/MainActivity.java
```diff
@@ -29,7 +29,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
         // Set up the action bar.
         final ActionBar actionBar = getActionBar();
-        assert actionBar != null;
+        if (actionBar == null) throw new NullPointerException("ActionBar has returned null!");
```
thisandagain: :+1:

thisandagain:
Awesome. Looks great. R+ :+1: :cat2: :cake: :boom:

PR #1543, opened 2015-04-23
import Router class

cadecairos:
Android studio suggested this to fix the build error. it appears to have worked.


cadecairos:
r?

ryanw-ny:
@cadecairos This is also fixed in PR #1542, it was a issue that happened with the migration of the old version of the router and the new one that was just pulled yesterday.

thisandagain:
Thanks @cadecairos. R+

PR #1547, opened 2015-04-23
adding jsxhint

gvn:


gvn:
This is just adding js(x)hint. We should do a follow-on ticket for jscs.

Runnable via `npm run lint`.

Also, I left the files @Pomax is working on alone to avoid merge conflicts, so maybe another follow on for that...

thisandagain:
:smile: Awesome. R+

Issue #1514, opened 2015-04-15
Implement map view front-end


xmatthewx:
Work started in #1504

flukeout:
Early feedback on @gvn's work:
* [ ] Start map in 'middle zoom' by default
* [ ] Slow down transition when adding a new tile. Add in the new tile and (+), then pan over
* [ ] Make map pannable on 6-up and 3-up views

gvn:
I just edited your previous comment. I think for *alpha* we should prioritize missing/broken functionality over aesthetic updates.

flukeout:
@gvn Sounds good, I'll use that comment above to add new items as I find them.

xmatthewx:
- [UI comps](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae0e362d5_2_52)
 - [UX notes](https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit#slide=id.gae931fbd6_7_20) (2 slides)

gvn:
@xmatthewx Are there unflattened versions of these comps available?

xmatthewx:
@vazquez has collected the icons. Let's share them via gDrive.

vazquez:
I have placed all Icons in `svg` format are available in Google Drive. 

WMA > Version 2.0 >  icons

@gvn Is this what you need?

gvn:
That's helpful, but I really appreciate a source/unflattened version of any comp I'm building so that I can get in and measure things more accurately.

gvn:
Also can you please just attach directly to the ticket? It's much easier for me to locate that way...

vazquez:
I can't upload Sketch files so here is the link: https://www.dropbox.com/s/alb2u2xroin6mmx/Map%20View.sketch?dl=0

I added a couple of extra screens just for context. I also added Icons as you will need them.

:rocket:

k88hudson:
Hey everyone, I spoke to @xmatthewx, @flukeout, and @vazquez about the flow for the editing view for the 6-up, 3-up, and 1-up zoom levels, I just wanted to post my summary here so we've got a unified record of what the flow is

https://drive.google.com/file/d/0B9Spnx3qAl7pamtRQTJfLUg2elE/view?usp=sharing

Let's discuss at stand up.

[...]

PR #1550, opened 2015-04-23
V1 refining

gvn:
Resolving functional issues from #1514


thisandagain:
Nice work on the panning ... feels good / smooth. R+ :point_up_2:

PR #1553, opened 2015-04-27
state binding, first pass

k88hudson:
First pass on binding react state to Android. I've hooked into the `componentWillMount` and `componentDidUpdate`. Could use some feedback before applying to the rest of the pages.

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
+++ b/app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
@@ -4,15 +4,36 @@ import mozilla.org.webmaker.router.Router;
 import android.content.Context;
 import android.util.Log;
 import android.webkit.JavascriptInterface;
+import android.content.SharedPreferences;
 
 import mozilla.org.webmaker.router.Router;
 
 public class WebAppInterface {
 
     private Context mContext;
+    private SharedPreferences mPrefs;
+    private String mPrefKey;
+    private String mPageState;
 
     public WebAppInterface(Context c) {
         mContext = c;
+        mPrefKey = mContext.getClass().getSimpleName() + "_prefs";
+        mPrefs = mContext.getSharedPreferences(mPrefKey, 0);
+        mPageState = mPrefs.getString("page_state", "{}");
+        Log.v("wm", "getting state " + mPrefKey + ": " + mPageState);
+    }
+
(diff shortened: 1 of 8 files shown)


k88hudson:
@thisandagain specifically I am looking for guidance on whether using `SharedPreferences`  the right thing to do, and whether namespacing serialized state objects with the activity name is the right approach:


```java
        mPrefKey = mContext.getClass().getSimpleName() + "_prefs";
        mPrefs = mContext.getSharedPreferences(mPrefKey, 0);
        mPageState = mPrefs.getString("page_state", "{}");
...

  @JavascriptInterface
    public void setState(String serializedState) {
        SharedPreferences.Editor edit = mPrefs.edit();
        edit.putString("page_state", serializedState);
        edit.commit();
        Log.v("wm", "set state to " + serializedState);
    }

```

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -4,15 +4,36 @@
...
 public class WebAppInterface {
 
     private Context mContext;
+    private SharedPreferences mPrefs;
+    private String mPrefKey;
+    private String mPageState;
 
     public WebAppInterface(Context c) {
         mContext = c;
+        mPrefKey = mContext.getClass().getSimpleName() + "_prefs";
```
thisandagain: Perhaps "_state" instead of "_prefs"?

thisandagain:
Having an NS to each saved state makes the most sense ... it seems likely that this type of persistence would cause inadvertent conflicts without that.

thisandagain:
Totally appropriate use of `SharedPreferences`. I'm wondering how you see this working when the individual page or component needs to retrieve JSON in `componentWillMount` from the API? Always persisting state by default is probably not a bad idea. :smile:

thisandagain:
R+

k88hudson:
We could have each activity fetch data from the API and override state as necessary; alternatively, we could have the JS make api calls, and resolve the overrides with the previous state itself.

thisandagain:
I'm pretty ok with either approach. The only thing that seems tricky is that Android has great support for offline, caching, and queuing network requests whereas we might need to build / find something that handles those cases gracefully in JS. The downside of using the Android layer is that the interface will be gross as callbacks across the JS / Java barrier are gross.

PR #1556, opened 2015-04-27
organizing less files

gvn:
An initial stab at LESS organization.

This puts LESS files parallel to their target view or component and names them with the same prefix.

My thought is that the `components` directory contains global/re-usable components and the `pages` (rename to `views`?) directory contains all view-specific LESS.


gvn:
How does this feel @thisandagain and @k88hudson ?

www_src/main.less
```diff
@@ -1,6 +1,17 @@
 @import "normalize";
-@import "project-map";
-@import "positionable";
+
+// Components
+
+@import "components/link/link.less";
```
k88hudson: .less not needed

gvn: Whoops...autocomplete.

k88hudson:
Looks good to me, a great improvement. A couple of notes

* We may want to have different entry point less files for each page eventually. I think @thisandagain was thinking that would be a better idea.  If we run into really big file sizes or something we could reconsider, but personally, I think your current approach simpler and works fine.
* There are a few more moving pieces/things that might be better automated (adding the right ID to the container div, adding the import statement), but happy to see those filed and done as follow-ups.

gvn:
Yeah, I was thinking I could document the standards once we agree on them, and then we can automate later.

k88hudson:
@gvn awesome, thanks for getting this done

thisandagain:
This is awesome. Thanks for the quick turnaround @gvn ! :panda_face:

gvn:
@Pomax FYI – Might wanna do a rebase first thing tomorrow if you have any patches you're working on!

PR #1572, opened 2015-04-27
matching header bar and zoom buttons to luke's design

gvn:


thisandagain:
Looks good. I'm really confused about the need for a back button and project title inside the app though ... @xmatthewx what is the canonical resource that we should be looking at for building this? If it is page 21 of the build document, then we can remove those header elements. That said, it raises another concern which is why is that back button on that comp different and what is the expected behavior?

https://docs.google.com/presentation/d/1ihVXe53q6U7YLKiYURKXedE5TzOW7x7iwCNeJBY4dpE/edit?usp=sharing

screenshot_2015-04-27-17-03-41

xmatthewx:
This looks like an artifact of the Map view when it was a standalone prototype.  

 - We do not need title and back arrow below the blue bar
 - We use standard back arrow for settings and other sibling views
 - We chose an exit button to exit projects
  - Projects have a parent child relationship with Discover
  - Exit may happen after a long immersive experience, making a simple back less intuitive.

In today's daily, @vazquez is going to discuss whether we will use Invision or the build deck as our canonical reference. 

tile map view tile selected

PR #1579, opened 2015-04-28
Editor view UI

k88hudson:
This is just a UI pass, as well as some components:

- Button
- ColorGroup
- Range

there are some performance issues with range around model updating


www_src/components/color-group/color-group.jsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  getInitialState: function () {
+    return {
+      value: 'transparent'
+    };
+  },
+  render: function () {
+    var linkState = this.props.linkState || this.linkState;
+    var valueLink = linkState(this.props.id);
+    function onChange(e) {
```
gvn: Seems weird to have this be in the render function. I'd make it a method instead...

www_src/components/range/range.jsx
```diff
@@ -0,0 +1,28 @@
...
+  },
+  getInitialState: function () {
+    return {
+      value: 100
+    };
+  },
+  render: function () {
+    var linkState = this.props.linkState || this.linkState;
+    var valueLink = linkState(this.props.id);
+    return (<div className="range">
```
gvn: Kind of a nit, but I find render returns to be more readable like this:

```javascript
return (
  <div className="range">
    <input min={this.props.min} max={this.props.max} type="range" valueLink={valueLink}/>
    <div className={'range-summary' + (parseInt(valueLink.value, 0) === this.props.min ? ' min' : '')}>{valueLink.value}{this.props.unit}</div>
  </div>
);
```

www_src/components/range/range.less
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  input[type=range]::-webkit-slider-thumb {
+    -webkit-appearance: none;
+    border: none;
+    height: 30px;
+    width: 30px;
+    border-radius: 36px;
+    background: #FFF;
+    cursor: pointer;
```
gvn: Is this for future desktop use?

www_src/static/img/change-image.svg
```diff
@@ -0,0 +1,16 @@
...
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
gvn: Recommend running this through `svgo` to optimize.

gvn:
Some nits/suggestions, but R+ :fire:

www_src/components/range/range.less
k88hudson: yeah idk, i figured why not

gvn:
Oh, but FWIW, I wasn't able to get this running on device. I saw a lot of "Webpage not available" messages.

PR #1588, opened 2015-04-28
Tinker

k88hudson:
Styles for the tinker page -- focus here is UI, I'm planning on working on hooking things up tmr.

Includes new components:

* ColorSpectrum
* Tabs


thisandagain:
:tada: :tada: :tada: :tada: :tada: 

Woah. This is pretty incredible progress. Really feels great overall. A few detailed UX/UI observations (no blockers, but things to track):
- Starting to scroll within the attribute and tinker mode views often leads to me accidentally dragging a slider
- The "down" state on each color cell is a highlighted square which breaks the illusion of "native-ness"
- I wonder if we might want to scale the image down and have the border width "expand" around the image until it hits 10px. Right now, when the image scales down to accommodate the border it feels a little odd
- The tab button down states are a bright blue highlight that covers only part of the tab area which feels a little odd. Do we have button / tab down states spec'd?

/cc @vazquez @xmatthewx

thisandagain:
Also realizing that we need to spec out the defaults. Right now I think transparency is defaulting to `70` for example. @xmatthewx do we already have any thoughts on this?

k88hudson:
Ok, I've removed blue highlights on buttons/inputs, fixed the image size in the preview. 

The way to prevent copy/paste is to set `user-select: none` on elements. I'll file that.

The scrolling interfering with sliding is going to be interesting to solve. I'm just using native ranges at the moment.

thisandagain:
Great. Thanks. R+ :panda_face: :cat:

PR #1611, opened 2015-04-30
better touch handling for project elements

Pomax:
This should be better touch handling for single and two finger gestures.

### DIFF
--- a/www_src/pages/project/Positionable.jsx
+++ b/www_src/pages/project/Positionable.jsx
@@ -36,7 +36,7 @@ var Positionable = React.createClass({
     var onode = this.refs.overlay.getDOMNode();
     onode.addEventListener("touchstart", touchHandler.secondFinger);
     onode.addEventListener("touchmove", touchHandler.panmove);
-    onode.addEventListener("touchend", touchHandler.endSecondFinger);
+    onode.addEventListener("touchend", touchHandler.endmark);
   },
 
   componentDidUpdate: function(prevProps, prevState) {
--- a/www_src/pages/project/Positionable.jsx
+++ b/www_src/pages/project/Positionable.jsx
@@ -80,8 +80,8 @@ var Positionable = React.createClass({
 
   handleTranslation: function(x, y) {
     this.setState({
-      x: /* this.transform.x + */ x,
-      y: /* this.transform.y + */ y
+      x: x,
+      y: y
     }, function() {
       if(this.props.onUpdate) {
         this.props.onUpdate(this.state);
@@ -91,8 +91,8 @@ var Positionable = React.createClass({
 
   handleRotationAndScale: function(angle, scale) {
     this.setState({
-      angle: /* this.transform.angle + */ angle,
-      scale: /* this.transform.scale * */ scale
+      angle: angle,
+      scale: scale
     }, function() {
       if(this.props.onUpdate) {
         this.props.onUpdate(this.state);
--- a/www_src/pages/project/Positionable.jsx
+++ b/www_src/pages/project/Positionable.jsx
@@ -20,18 +20,22 @@ var Positionable = React.createClass({
   componentDidMount: function() {
     if(!this.state.interactive) return;
     var touchHandler = this.touchhandler = require("./touchhandler")(this);
-
     var dnode = this.getDOMNode();
-    dnode.addEventListener("touchstart", touchHandler.startmark);
     dnode.addEventListener("mousedown", touchHandler.startmark);
-    dnode.addEventListener("touchmove", touchHandler.panmove);
     dnode.addEventListener("mousemove", touchHandler.panmove);
-    dnode.addEventListener("touchend", touchHandler.endmark);
     dnode.addEventListener("mouseup", touchHandler.endmark);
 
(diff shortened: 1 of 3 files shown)


www_src/pages/project/touchhandler.js
```diff
@@ -1,69 +1,126 @@
...
+     */
     startmark: function(evt) {
-      positionable.transform = JSON.parse(JSON.stringify(positionable.state));
-      positionable.transform.markx = evt.clientX || evt.touches[0].pageX;
-      positionable.transform.marky = evt.clientY || evt.touches[0].pageY;
-      positionable.setState({ touchactive: true });
-      if(evt.touches && evt.touches.length > 1) { secondFinger(evt); }
       evt.preventDefault();
       evt.stopPropagation();
+      console.log("mark!")
```
thisandagain: stray `console.log` here.

www_src/pages/project/touchhandler.js
```diff
@@ -1,69 +1,126 @@
...
+      if(evt.touches.length === 1) {
+        mark = JSON.parse(JSON.stringify(positionable.state));
+        console.log(JSON.stringify(mark));
```
thisandagain: another stray `console.log`

www_src/pages/project/touchhandler.js
```diff
@@ -1,69 +1,126 @@
...
+     */
     panmove: function(evt) {
-      if (!positionable.state.touchactive) return;
+      evt.preventDefault();
+      evt.stopPropagation();
       if (evt.touches && evt.touches.length > 1) {
-        return this.secondFinger(evt);
+        return handlers.handleTouchRepositioning(evt);
       }
+      console.log("panmove");
```
thisandagain: `console.log`

thisandagain:
Performance feels great. Hard to describe the STR for what I'm seeing during testing but I'll try my best:

#### Setup
- Open the page editor
- Select an element with one finger and start to drag it (easiest with Tucker)
- With another finger start to scale and rotate the element
- Release the second finger and then tap with it all over the page view

#### Expected result
- Nothing

#### Actual result
- The selected element will start to "jump" in rotation, position, and scale. In some cases doing this repeatedly can cause the element to seemingly disappear from the view.

thisandagain:
FWIW I'm able to cause a similar issue in the current build as well, just much more difficult to STR.

www_src/pages/project/touchhandler.js
thisandagain: I'm going to stop noting them. :smile:

www_src/pages/project/Positionable.jsx
```diff
@@ -44,7 +48,7 @@ var Positionable = React.createClass({
   render: function() {
     var x = this.state.x,
         y = this.state.y,
-        angle = this.state.angle,
+        angle = this.state.angle * 180/Math.PI,
```
thisandagain: :+1:

www_src/pages/project/touchhandler.js
```diff
@@ -1,69 +1,126 @@
...
+      y1: false,
+      x2: false,
+      y2: false,
+      distance: 0,
+      angle: 0
+    };
+  };
+
+  var transform = resetTransform();
+  var mark = JSON.parse(JSON.stringify(positionable.state));
```
thisandagain: I've started to bring in some single serving utilities ... perhaps we could include [lodash.clone](https://www.npmjs.com/package/lodash.clone) to make this a little easier to read?

k88hudson: Use Object.assign from react -- 
```
var assign = require('react/lib/Object.assign');
...
var whatever = assign({}, props, {foo: 'bar'});
```

Pomax:
hm, sounds like the multi-finger-release is not properly kicking in, then. should be debuggable by giving the super sneaky squirrel overlay a background color and seeing if it gets unassigned or stays active. I shall investigate

www_src/pages/project/touchhandler.js
Pomax: not in this case. Assign preserves references, and in this case I need guaranteed "these are new values, completely detached from any and all preexisting bindings".

k88hudson: Ah ok I figured a shallow copy would be ok here since your state object seems to only have primitive values but this is cool too

gvn: Also, this is my favorite object copy hack ever. Just sayin'. :rabbit:

thisandagain: Got it. :+1:

k88hudson: :sheep: <-- this is officially my weird sheep now

Pomax: fun fact about this kind of deep copying: it might seem like a hack, but is actually the "normal" way to force a deep copy in many languages that predate JS. In Java, for instance, a deep copy in absence of copy constructors is "serialize the instance to bytecode, then read it back in using the same class formats", and done. Instant deep copy.

thisandagain:
R+ :shipit: Some minor adjustments still to do, but most are UX so we can file individual tickets. :smile:

Issue #1596, opened 2015-04-29
Create a colors less file


k88hudson:
Because it makes us happy

k88hudson:
cc @gvn

gvn:
@vazquez - Do you have a standardized palette that you're working from?

gvn:
Also, this palette should be represented in our built style guide.

vazquez:
@gvn Will have something for you by the end of the day. Sorry about the delay.

vazquez:
![colours](https://cloud.githubusercontent.com/assets/1119821/7435791/f474d99a-f014-11e4-854d-1fb963e5d56d.png)

This is the Sketch file: https://www.dropbox.com/s/fivy1fgmq5etiis/Colours.sketch?dl=0

@gvn Let me know if I'm missing any as we go along.

gvn:
That's super awesome. I will add it to the style guide!

gvn:
I think we should name the colors what they are instead of how they will be used since they'll undoubtedly be used for more than one purpose. I'll name the LESS variables in the build after those colors...

EG: "Light Blue", "Heather Gray", etc.

@vazquez Do you want to name them?

vazquez:
Oh boy! :smile: Okay give me 5 minutes. Have to harness my inner colour connoisseur.

gvn:
Definitely need a "Vivid Tangerine". :wink:

vazquez:
Hopefully I didn't go too crazy. I provided a more common option in parentheses for a couple of them.

![colours](https://cloud.githubusercontent.com/assets/1119821/7436059/ab6d8bb8-f017-11e4-9c2b-786cc7c33c61.png)

gvn:
That's perfect! I'll add them!

thisandagain:
:heart: "Shadow Plum". I'll update the Android `colors.xml` to match as well.

k88hudson:
beautiful :dancers:

davidascher:
I think the light grey is more of a "dusty fog" myself.

(spent too much time looking at paint chips w/ Sylvia)
​

gvn:
Adding colors in #1627.

Will still need to replace all the hard coded values in the build once it lands...

PR #1617, opened 2015-04-30
Cartesian map stuff

k88hudson:


k88hudson:
@gvn R?

www_src/components/action-menu/action-menu.jsx
```diff
@@ -0,0 +1,49 @@
...
+    var className = classNames(this.props.className, {
+      'primary-btn': true,
+      'off': this.props.off
+    });
+    var props = assign({}, this.props, {className});
+    return <Btn {...props} />;
+  }
+});
+
+var Secondary = React.createClass({
```
gvn: This seems almost identical to Primary. Any way they can be combined or derived from a base class?

www_src/components/action-menu/action-menu.less
```diff
@@ -0,0 +1,81 @@
+/*********************************************************
+* Action Menu
+*/
+.action-menu {
+  @_bottomGutter: 10px;
+
+  @_primaryBg: #69A0FC;
+  @_primaryText: #FFF;
+  @_secondaryBg: #FFF;
+  @_secondaryText: #718095;
```
gvn: I'm thinking we're gonna probably need a colors.less and some kind of standardized palette soon...

www_src/components/action-menu/action-menu.jsx
k88hudson: Oh yeah... these totally ended up the same, haha.

www_src/components/action-menu/action-menu.less
k88hudson: Yeah i filed https://github.com/mozilla/webmaker-app/issues/1596

k88hudson: For now i've been putting them up top with underscores so we can replace them easily later

www_src/components/action-menu/action-menu.less
```diff
@@ -0,0 +1,81 @@
...
+  @_primarySize: 56px;
+  @_secondarySize: 40px;
+  @_spacing: 56px;
+
+  position: fixed;
+  left: 50%;
+  bottom: @_bottomGutter;
+  z-index: 999;
+  width: @_primarySize;
+  margin-left: -(@_primarySize/2);
```
gvn: Can this not just be centered with flexbox?

www_src/components/action-menu/action-menu.less
```diff
@@ -0,0 +1,81 @@
...
+  @_secondaryText: #718095;
+
+  @_primarySize: 56px;
+  @_secondarySize: 40px;
+  @_spacing: 56px;
+
+  position: fixed;
+  left: 50%;
+  bottom: @_bottomGutter;
+  z-index: 999;
```
gvn: Probably only needs like `2`.

www_src/components/action-menu/action-menu.less
```diff
@@ -0,0 +1,81 @@
...
+    color: @_secondaryText;
+    &.left {
+      margin-left: -(@_secondarySize/2 + @_spacing);
+    }
+    &.right {
+      margin-left: -(@_secondarySize/2 - @_spacing);
+    }
+    &.off {
+      transform: translateY(@_bottomGutter + @_secondarySize * 2);
+    }
```
gvn: All this calculation based on fixed pixel values makes me cringe a bit, but I guess it's fine...

www_src/components/tabs/tabs.jsx
```diff
@@ -1,5 +1,5 @@
 var React = require('react');
-var classNames = require('classNames');
+var classNames = require('classnames');
```
gvn: Oh, :camel:

www_src/components/action-menu/action-menu.less
k88hudson: yeah probably :palm_tree:

www_src/components/action-menu/action-menu.less
k88hudson: in this case i think it's better to do this like this since using flexbox would require a large container element at full width, which will block content behind it

www_src/components/action-menu/action-menu.less
k88hudson: i could alternatively use calc (gutter has to be fixed pixel) but i feel like there's not really a downside to using the pixel calculation since we need exact pixel sizes for these buttons

www_src/lib/cartesian.js
```diff
@@ -0,0 +1,129 @@
...
+      paddingTop: this.getAbsDistance('y', 'min'),
+      paddingRight: this.getAbsDistance('x', 'max'),
+      paddingBottom: this.getAbsDistance('y', 'max'),
+      paddingLeft: this.getAbsDistance('x', 'min'),
+      marginLeft: -((diffX * this.widthPadded + this.width) / 2) + 'px',
+      marginTop: -((diffY * this.heightPadded + this.height) / 2) + 'px',
+    };
+  }
+
+  getFocusTransform(coords) {
```
gvn: Might want to comment this function

www_src/pages/map/grid.jsx
```diff
@@ -1,440 +0,0 @@
...
-      }
-    }
-
-    return (
-      <div className="grid">
-        {nodes}
-      </div>
-    );
-  }
-});
```
gvn: :sob:

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
+  },
+  componentDidMount: function () {
+    var el = this.getDOMNode();
+    var bounding = this.refs.bounding;
+    var boundingEl = bounding.getDOMNode();
+    var startX, startY, deltaX, deltaY;
+    var didMove = false;
+
+    // Set up initial translation
+    this.setState({
```
gvn: Could this be set in getInitialState instead? I think this means render will happen 2x in a row...

gvn:
When you're fully zoomed out and delete a tile, the map zooms in. Is that intentional UX?

www_src/pages/map/map.jsx
gvn: Not that that's a huge deal, but ༼ つ ◕_◕ ༽つ

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
-    //   since container must be measured before tiles can be properly laid out...
-    var elWrapper = this.refs.wrapper.getDOMNode();
-    this.refs.masterGrid.setContainerDimensions(elWrapper.clientWidth, elWrapper.clientHeight);
-
-    setTimeout(function() {
-      this.refs.masterGrid.setZoomLevel(this.state.zoomLevel);
-    }.bind(this), 100);
+  addPage: function (coords) {
+    return () => {
+      var id = 'foo' + elements.length
```
gvn: LOL. Uhm...can we add a GUID lib or something?

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
+      this.setState({
+        camera: this.cartesian.getFocusTransform(coords),
+        selectedEl: id
+      });
+    };
   },
-  onZoomChange: function (event) {
+  removePage: function () {
+    var index;
+    elements.forEach((el, i) => {
```
gvn: This is kinda why I feel a little weird about this new schema. I almost wonder if we should have our keys just be IDs in an object instead of using an array. Having to iterate over the array to lookup by ID seems not especially performant. I don't think order matters either?

Like:

```javascript
{
  'guid12345': {
    coords: {x: 0, y: 0},
    text: 'Hello'
  },
  'guid52353': {
    coords: {x: 1, y: 0},
    text: 'Bye'
  }
}
```

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
+
+    var addContainerStyle = classNames('page-container add', {off: this.state.zoom == 1});
+
     return (
-      <div id="map" className={ 'zoom-' + this.state.zoomLevel }>
-        <div ref="wrapper" className="wrapper">
-          <div>
-            <Grid ref="masterGrid" aspectRatio={"35:40"} onZoomChange={ this.onZoomChange } />
+      <div id="map">
+
```
gvn: Nit: unnecessary blank line?

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
-            onTap={ this.zoomGridIn }></Hammer>
-          <Hammer
-            className="delete"
-            component="button"></Hammer>
-        </div>
+        <Menu>
+          <Secondary side="left" off={this.state.zoom <= 0.25} onClick={this.zoomOut} icon="../../img/zoom-out.svg" />
+          <Secondary side="right" off={!this.state.selectedEl} onClick={this.removePage} icon="../../img/trash.svg" />
+          <Primary onClick={this.zoomIn} off={this.state.zoom >= 1} icon="../../img/zoom-in.svg" />
+          <Primary url="/projects/123/elements/1" href="/pages/project" off={this.state.zoom < 1} icon="../../img/pencil.svg" />
```
gvn: As discussed: `PrimaryButton` and `SecondaryButton` are more semantic.

www_src/pages/map/map.jsx
k88hudson: hm. that's a good idea, maybe drop that in the API discussion?

www_src/pages/map/map.jsx
```diff
@@ -1,66 +1,170 @@
...
+        </Menu>
+
       </div>
     );
   }
 });
 
-render(App);
+// Render!
```
gvn: Could probably do without this comment... ;)

www_src/pages/map/map.jsx
k88hudson: aw ok -_-

www_src/pages/map/map.jsx
k88hudson: hm. i suppose we could if we create `this.cartesian` in getInitialstate actually

www_src/components/action-menu/action-menu.less
gvn: Hmm...good point.

www_src/pages/map/grid.jsx
k88hudson: :heart:

www_src/components/action-menu/action-menu.less
gvn: Yeah it's fine...

www_src/pages/map/grid.jsx
k88hudson: *hug*

www_src/lib/cartesian.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+  get widthPadded() {
+    return this.width + this.gutter;
+  }
+  get heightPadded() {
+    return this.height + this.gutter;
+  }
+
+  // Calculate absolute distance from 0 to min/max
+  getAbsDistance(xOrY, maxOrMin) {
```
gvn: I'm having a hard time grokking this function, but could be my geometry being rusty...

www_src/pages/map/map.jsx
gvn: CC: @thisandagain

www_src/lib/cartesian.js
k88hudson: Sorry it's a bit confusing.

For the bounding box, in order to calculate how much padding to put on either side of the 0 case, I need the absolute distance to the lowest value (padding-right, padding-bottom) and the absolute distance to the highest value, for both the set of x and the set of y.

saves me having to do
```
 Math.abs(Math.min(allXValues));
 Math.abs(Math.max(allXValues));
 Math.abs(Math.min(allYValues));
 Math.abs(Math.max(allYValues));
```

(the apply is just so i can turn an array of values into arguments)

www_src/pages/map/map.jsx
k88hudson: I still need to maintain an array anyway to build the grid tho

gvn: I guess you could also try `find` instead of `forEach`.

gvn:
This is super awesome. Descartes wins again!

I think the foo-GUID is the only thing I'm really concerned with, but that can totally be a new ticket.

thumbs

PS: Please squash. :wink:

www_src/pages/map/map.jsx
thisandagain: Bring back that weird one-liner GUID lib!? :stuck_out_tongue_winking_eye:

gvn: `(Date.now() + '' + Math.random()).replace('.', '-')`

J/K

www_src/pages/map/map.jsx
k88hudson: yeah this was actually a way better way to do it. good call!

k88hudson:
Ok! the uuid lib has been resurrected, and all is well.

www_src/pages/map/map.jsx
gvn: Thanks :)

PR #1618, opened 2015-04-30
fixing js(x)hinting

gvn:


gvn:
The globbing for jsxhint is buggy, so this isn't hitting deeply nested js/jsx, but this is better than before...

k88hudson:
Looks great to me!

PR #1622, opened 2015-05-01
Discover View

thisandagain:


www_src/pages/discover/discover.jsx
```diff
@@ -1,42 +1,50 @@
...
   render: function () {
-    return (
-      <div id="discover" className="demo">
-        <Link url="/map/123" href="/pages/map" className="tile tile-demo">
-          <div className="inner">
-            <div className="img">
-              <img src="../../img/toucan.svg" />
-            </div>
+    var cards = [];
+    var project = null;
```
Pomax: `var project;` ?

www_src/pages/discover/discover.jsx
```diff
@@ -1,42 +1,50 @@
...
+    
+    for (var i = 0; i < this.state.list.length; i++) {
+      project = this.state.list[i];
+      console.dir(project.thumbnail[480]);
```
Pomax: console call

www_src/pages/discover/discover.jsx
```diff
@@ -1,42 +1,50 @@
...
+      cards.push(
```
Pomax: this is a "bad plan" in terms of React. For React, you want to use `map` with key attributes instead:
```
cards = this.state.list.map( project => {
  var key = "something that uniquely identifies this project";
  return <Link ... key={key} ...> ... </Link>;
});
```

www_src/pages/discover/discover.jsx
```diff
@@ -1,42 +1,50 @@
...
+            <div className="title">{project.title}</div>
+            <div className="author">{project.author.username}</div>
           </div>
         </Link>
-        <p><button onClick={this.onClick}>{this.state.test ? 'on' : 'off'}</button></p>
+      );
+    }
+
+    return (
+      <div id="discover">
```
Pomax: ideally, we don't rely on `id` anymore, since React components don't come with a uniqueness guarantee. Can we say something like `className="discover page"` instead, and then tap into that as CSS namespace?

www_src/pages/discover/discover.less
```diff
@@ -1,3 +1,46 @@
...
+        position: relative;
+
+        width: 97%;
+        margin: 0 auto;
+        background: #efefef;
+      }
+    }
+
+    .meta {
+      width: 97%;
```
Pomax: why 97? (mostly asking because anything that isn't 50 or 100 is probably a "this can be make perfect with `calc`" instance)

www_src/pages/discover/discover.less
```diff
@@ -1,3 +1,46 @@
...
+      margin: 6px auto;
+      padding: 10px 0;
+      font-size: 26px;
+    }
+
+    .author {
+      color: #00f;
+      font-size: 22px;
```
Pomax: can we make this a `rem` value instead?

www_src/pages/discover/mock.json
```diff
@@ -0,0 +1,126 @@
+[
+    {
+        "id": 1,
+        "version": "0.0.0",
```
Pomax: do semvers make sense for what is essentially simple version control? This feels like it should just be a plain revision number.

www_src/pages/discover/mock.json
```diff
@@ -0,0 +1,126 @@
...
+        "title": "Yam Fries",
+        "featured": true,
```
Pomax: No description? Might be valuable information to have for the discover view.

www_src/pages/discover/mock.json
```diff
@@ -0,0 +1,126 @@
...
+            "updated_at": "2015-04-20T16:11:13.961Z",
+            "deleted_at": null
+        },
+        "thumbnail": {
+            "480": "https://webmaker-app-mock.s3.amazonaws.com/tucker_000.jpg"
+        },
+        "author": {
+            "id": 1234,
+            "username": "billmurray",
+            "locale": {
```
Pomax: this feels like being specific in a way that is going to break the moment a country stops existing. Which has kind of happend *lots* of times in the last 20 years. Just keeping the locale a single code string feels much safer.

www_src/pages/discover/discover.jsx
thisandagain: Awesome. Will do. :+1:

www_src/pages/discover/discover.jsx
thisandagain: Yup! Sounds good. I was following the pattern we were using in a few other views but totally agree.

www_src/pages/discover/discover.less
thisandagain: :+1:

www_src/pages/discover/discover.less
thisandagain: Yup! Typography still needs a lot of work.

www_src/pages/discover/mock.json
thisandagain: Not included in design spec.

www_src/pages/discover/mock.json
thisandagain: Not being used at the moment, but included to keep consistency with @cadecairos API design. I'll keep as to maintain some continuity but we have lots of opportunities to modify this before launch.

www_src/pages/discover/mock.json
thisandagain: :+1: to maintaining a version history. Not included in v1 but something I'd like to consider in the future. Agreed to simplifying the version to just be a number that we increment.

www_src/pages/discover/mock.json
Pomax: :pitchforks:

www_src/pages/discover/mock.json
Pomax: hmm, this feels like the API design might be too specific, in that case. Worth looking at there.

package.json
```diff
@@ -64,6 +60,8 @@
     "lodash.defaults": "3.1.1",
     "react": "0.13.1",
     "react-hammerjs": "0.2.2",
+    "react-imageloader": "1.2.0",
+    "react-loadermixin": "1.0.4",
```
Pomax: not used?

package.json
```diff
@@ -64,6 +60,8 @@
     "lodash.defaults": "3.1.1",
     "react": "0.13.1",
     "react-hammerjs": "0.2.2",
+    "react-imageloader": "1.2.0",
```
Pomax: let's also add a short section in the README.md here along the lines of "dependencies for common tasks" so people don't reinvent the wheel all the time, and have a quick place to find out what we use for which purpose

www_src/pages/discover/discover.jsx
```diff
@@ -1,42 +1,43 @@
...
+    var cards = this.state.list.map( project => {
+      return (
+        <Link url={"/map/" + project.id} href="/pages/map" key={project.id} className="card">
+          <div className="thumbnail">
+            <ImageLoader src={project.thumbnail[480]}>
+              // @todo Show image error icon / graphic
+            </ImageLoader>
+          </div>
+
+          <div className="meta">
```
Pomax: might be worth turning this into a `<Metadata>` component. Right now we're using title and author only, but we probably want to extend that later on with additional things like date posted, likes, etc.

package.json
thisandagain: `ImageLoader` is used in `discover.jsx`. `LoaderMixin` is a dependency of the image loader.

www_src/pages/discover/discover.jsx
```diff
@@ -1,43 +1,43 @@
...
+              // @todo Show image error icon / graphic
+            </ImageLoader>
+          </div>
 
-        <ProjectSnapshot
-          url="/map/123"
-          href="/pages/map"
-          thumbnail="../../img/toucan.svg"
-          title="More birds"/>
+          <div className="meta">
```
Pomax: might be worth creating a `<Metadata>` component here. Right now it's only author/title, but I can see date posted, views, etc. also ending up in there.

package.json
thisandagain: Totally agree. :+1:

package.json
Pomax: That's imageloader's dependency, no ours; it should be required in the react-imageloader package.json, which looking at https://github.com/hzdg/react-imageloader/blob/master/package.json#L38 seems to be the case?

package.json
thisandagain: Added.

README.md
```diff
@@ -58,8 +65,23 @@ Component markup should contain a top-level class name that corresponds to its f
...
+api({
+    uri: '/discover',
+    useCache: false
+}, function (err, results) {
+    // do stuff! 
+});
+``` 
 
+#### Loading Images
+Any time you are loading images over the network, we recommend that you use the `ImageLoader` react component. This gives you access to important events like loading and error states as well as a hook for providing a loading animation. Full documentation can be found here: https://github.com/hzdg/react-imageloader
```
Pomax: small nit: let's use `<ImageLoader>` to make it clear that it's not a library or the like, but an actual component

www_src/pages/discover/discover.jsx
Pomax: filed https://github.com/mozilla/webmaker-app/issues/1656 as followup

Issue #1626, opened 2015-05-01
Add built style guide


gvn:
Let's create a style guide view in the build that represents all re-usable components, styles, colors, etc.

PR #1627, opened 2015-05-01
starting a built style guide

gvn:
Related to #1626


gvn:
This is starting to expose some of the improvements we can make to our reusable components so they work in broader contexts.

(eg: Make Range expand to full width. Remove padding on Tabs and add it contextually.)

My hope is that we can start developing components *in* the style guide view first and then place them into more specific contexts. I've taken that approach on many projects with great results!

CC: @thisandagain @Pomax

gvn:
@xmatthewx I figure this will interest you as well... :dancer: 

styleguide

thisandagain:
:goat: :smile: :panda_face: :cat:

xmatthewx:
heck yes

@vazquez - this can serve as a visual test page to see if future changes funk anything up.

gvn:
Yep, and it'll definitely be a living, growing, and hopefully canonical document!

k88hudson:
Sweet this is awesome. Ring now

k88hudson:
personally i would deemphasize the colour names more since the dark background kinda affects the way you see the adjacent colour but this is awesome, merge away!

PR #1633, opened 2015-05-01
Text editor

k88hudson:


k88hudson:
Missing:

* actual text editing (i.e. content editable). that needs to be addressed in a patch of its own
* white states for icons. because i don't feel like doing another set of white ones ok
* actual custom font loading. needs its own ticket.

www_src/static/img/B.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
Pomax: all these SVG have an aweful lot of "sketch" branding... I'd be highly in favour of stripping that away.

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,82 @@
+var React = require('react/addons');
+
+// Internal -- just the container
+var OptionPanel = React.createClass({
```
Pomax: it might be worth doing these components as one component per file and making this a plain js `module.exports = { require("checkbox"), ..., ... }`

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,82 @@
...
+      truthy: true,
+      falsey: false
+    };
+  },
+  onChange: function (e) {
+    var val = e.target.checked ? this.props.truthy : this.props.falsey;
+    this.valueLink.requestChange(val);
+  },
+  isTruthy: function () {
+    return this.valueLink.value === this.props.truthy;
```
Pomax: given that this isn't true or false, but truthy and falsey, should this be `==` rather than `===`?

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,82 @@
...
+  },
+  render: function () {
+    var linkState = this.props.linkState;
+    this.valueLink = linkState(this.props.id);
+    return (<label className={'label ' + (this.isTruthy() ? 'selected' : '')}>
```
Pomax: if you return `false` instead of `''`, className won't even be bound as property (so you don't end up with a `class=""` in the final DOM

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,82 @@
...
+});
+
+var CheckboxSet = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    linkState: this.linkState
+  },
+  render: function () {
+    return (<OptionPanel>
+      {this.props.options.map(props => <Checkbox {...props} linkState={this.props.linkState} />)}
```
Pomax: this is probably going to complain about missing `key` attributes

www_src/components/option-panel/option-panel.jsx
k88hudson: sorry i should name these something else

www_src/components/option-panel/option-panel.jsx
k88hudson: hm i'd rather not, they are pretty small

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,82 @@
...
+  onChange: function (e) {
+    console.log(e.target.value);
+    this.valueLink.requestChange(e.target.value);
+  },
+  render: function () {
+    this.valueLink = this.props.linkState(this.props.id);
+    var currentVal = this.valueLink.value;
+    return (<OptionPanel>
+      {this.props.options.map(item => {
+        return (<label className={'label ' + (item.value === currentVal ? 'selected' : '')}>
```
Pomax: missing `key` attributes? (fairly necessary if we want to take advantage of React's virtual DOM updating)

www_src/components/option-panel/option-panel.jsx
k88hudson: yeah good call

www_src/components/option-panel/option-panel.less
```diff
@@ -0,0 +1,33 @@
...
+      box-shadow: 0 2px 0 0 #4881DF;
+      border-left: 1px solid fade(#4881DF, 50%);
+    }
+    &:first-child {
+      border-radius: 3px 0 0 3px;
+    }
+    &:last-child {
+      border-radius: 0 3px 3px 0;
+    }
+    input {
```
Pomax: What's the difference between this pair of props and `display:none`?

www_src/components/select/select.less
```diff
@@ -0,0 +1,12 @@
+.select {
+  -webkit-appearance: none;
```
Pomax: I seem to recall we did auto-prefixing, so the -moz- and -webkit- prefixes should be unnecessary

www_src/static/img/B.svg
gvn: This reminds me, I need to add a `svgo` task.

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
   }
 });
 
+var TextEditor = React.createClass({
```
Pomax: given the size of this component, definitely worth putting in its own file.

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
...
+  mixins: [React.addons.LinkedStateMixin, Binding],
+  getInitialState: function () {
+    return {
+      fontSize: 18,
```
Pomax: `1rem` instead?

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
...
+      color: '#645839',
+      fontWeight: 'normal',
+      fontStyle: 'normal',
+      textDecoration: 'none',
+      textAlign: 'center'
+    };
+  },
+  onChange: function (id) {
+    return (e) => {
+      var state = {};
```
Pomax: If we're clearing and then setting a single property, we could do that directly as a `var state = { id: e.target.checked };`, although state does not "clear" with setState, it simply updates any keys provided, so in this case it'd just be `this.setState({ id: e.target.checked });`

www_src/components/select/select.less
k88hudson: unfortunately autoprefixer doesn't autoprefix "appearance" beacause it's technically non standard

www_src/pages/editor/editor.jsx
k88hudson: no

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
...
+      },
+      {
+        value: 'right',
+        icon: '../../img/align-right.svg',
+      }
+    ];
+    return (
+      <div id="editor">
+        <div className="editor-preview">
+          <p style={this.state}>Hello world</p>
```
Pomax: this feels like a dangerous, or at least odd, thing to do. Why would the entire state of the text editor be equivalent to CSS styling?

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
...
+        </div>
+        <div className="editor-options">
+          <div className="form-group">
+            <label>Text Size</label>
+            <Range id="fontSize" min={10} max={32} unit="px" linkState={this.linkState} />
+          </div>
+          <div className="form-group">
+            <label>Font</label>
+            <select className="select" valueLink={this.linkState('fontFamily')}>
+              <option value="Roboto">Roboto</option>
```
Pomax: three options with `value` equal to `innerText` feels like it can use an array.map somehow

www_src/pages/editor/editor.jsx
```diff
@@ -48,5 +48,96 @@ var Editor = React.createClass({
...
+          </div>
+        </div>
+      </div>
+    );
+  }
+});
+
 // Render!
-render(Editor);
+render(window.location.hash === '#text' ? TextEditor : ImageEditor);
```
Pomax: not too familiar with our options here, is this the best way to switch which view to serve up?

www_src/components/select/select.less
Pomax: oh weird. do we control the "list of properties to prefix"? (that feels like a thing we should be able to control"

www_src/pages/editor/editor.jsx
Pomax: why? (just thinking in terms of accessibility-without-putting-in-effort here)

www_src/pages/editor/editor.jsx
k88hudson: `state[id]` not state['id']`

www_src/pages/editor/editor.jsx
k88hudson: because i'm lazy

www_src/pages/editor/editor.jsx
Pomax: I am teh blindz. Sometimes, **infrequently**, but sometimes, I wish JS had PHP's $$var concept (where {$$id: ... } would be the same as obj={}, obj[id] = ... )

www_src/pages/editor/editor.jsx
k88hudson: temporary hack tbh.

there are a number of ways we can do this. probably the best way is to pass the params from the android router, and then render a child component based on what the type of the element is.

im planning on working on hooking this stuff up to real api responses etc next week.

www_src/pages/editor/editor.jsx
k88hudson: ah you know what, that reminds me, i could totally have used an es6 symbol here! although actually this function isn't being used  soooo i can just remove it, heh

www_src/components/option-panel/option-panel.jsx
k88hudson: in this case you always have `label`, it's just the second class that gets added or not.

k88hudson: i guess i should just do this with classNames

k88hudson:
@Pomax ok, review changes in

www_src/main.less
```diff
@@ -56,6 +58,25 @@ code, pre {
   }
 }
 
+// Screen readers only
```
Pomax: little confused by the comment plus actual rules: screen readers only really care about display:none or visibility:hidden, so it seems that we don't need all these rules. A simple "display:none" should already achieve the goal. But then the question becomes: if sr-only is always active for these elements, how does this do different things for screen readers vs. not screen readers?

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+});
+
+var Checkbox = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    return {
+      linkState: this.linkState,
+      checkedValue: true,
+      uncheckedValue: false
```
Pomax: why do we need two states for this? uncheckedValue should *always* be !checkedValue

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+  },
+  onChange: function (e) {
+    var val = e.target.checked ? this.props.checkedValue : this.props.uncheckedValue;
+    this.valueLink.requestChange(val);
+  },
+  isChecked: function () {
+    return this.valueLink.value === this.props.checkedValue;
```
Pomax: should this be `this.state.checkedValue`?

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+    </div>);
+  }
+});
+
+var Checkbox = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    return {
+      linkState: this.linkState,
+      checkedValue: true,
```
Pomax: if we can pass this value down through props, this should probably be `checkedValue: this.props.checkedValue` instead

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+  onChange: function (e) {
+    var val = e.target.checked ? this.props.checkedValue : this.props.uncheckedValue;
+    this.valueLink.requestChange(val);
+  },
+  isChecked: function () {
+    return this.valueLink.value === this.props.checkedValue;
+  },
+  render: function () {
+    var linkState = this.props.linkState;
+    this.valueLink = linkState(this.props.id);
```
Pomax: does this need to be rebound on every `render`? Is this something we can simply do once in `componentWillMount` instead?

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+});
+
+var Radio = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    linkState: this.linkState
+  },
+  onChange: function (e) {
+    console.log(e.target.value);
```
Pomax: rogue console log

www_src/components/select/select.less
Pomax: small bump - if we can append to the list of "to prefix" elements, that sounds like worth doing here.

www_src/components/option-panel/option-panel.jsx
k88hudson: this is `getDefaultProps`, what do you mean?

www_src/pages/editor/editor.jsx
```diff
@@ -1,52 +1,8 @@
...
-      </div>
-    );
-  }
-});
+var ImageEditor = require('./image-editor.jsx');
+var TextEditor = require('./text-editor.jsx');
 
 // Render!
-render(Editor);
+render(window.location.hash === '#text' ? TextEditor : ImageEditor);
```
Pomax: it feels like this is a thing we should be able to query React on, instead of relying on "window".

www_src/components/option-panel/option-panel.jsx
k88hudson: these are default values. you could also set the checked state to be `'off'` and `'on'`, or whatever else you want

www_src/pages/editor/image-editor.jsx
```diff
@@ -0,0 +1,49 @@
...
+      transparency: 70,
+      borderWidth: 0,
+      borderColor: 'transparent'
+    };
+  },
+  render: function () {
+    var style = {
+      opacity: this.state.transparency / 100,
+      borderStyle: 'solid',
+      borderWidth: this.state.borderWidth,
```
Pomax: currently set to 0, but if it ever gets a real value, this will break due to a missing unit. Do we want to suffix "px" anyway?

www_src/pages/editor/text-editor.jsx
```diff
@@ -0,0 +1,93 @@
...
+      fontFamily: 'Roboto',
+      color: '#645839',
+      fontWeight: 'normal',
+      fontStyle: 'normal',
+      textDecoration: 'none',
+      textAlign: 'center'
+    };
+  },
+  render: function () {
+    var textStyleOptions = [
```
Pomax: this feels like a thing we might use in more than one place. Maybe not something to refactor in this PR, but if not, should we file a followup to take this out and make it its own thing?

www_src/pages/editor/text-editor.jsx
```diff
@@ -0,0 +1,93 @@
...
+        checkedValue: 'italic'
+      },
+      {
+        id: 'textDecoration',
+        icon: '../../img/U.svg',
+        uncheckedValue: 'none',
+        checkedValue: 'underline'
+      }
+    ];
+    var textAlignOptions = [
```
Pomax: same question as above.

www_src/main.less
k88hudson: Read http://getbootstrap.com/css/#helper-classes-screen-readers

www_src/pages/editor/text-editor.jsx
```diff
@@ -0,0 +1,93 @@
...
+      },
+      {
+        value: 'right',
+        icon: '../../img/align-right.svg',
+      }
+    ];
+    return (
+      <div id="editor">
+        <div className="editor-preview">
+          <p style={this.state}>Hello world</p>
```
Pomax: this still feels like it needs to beunlazy'd - for instance, right now fontSize will be an illegal value, because it's a number without a unit.

www_src/components/option-panel/option-panel.jsx
k88hudson: no, this is a comparison of the state (this.valueLink.value) and the checkedValue.

so if your values map to `on` and `off`, you want to see if the state (this.valueLink.value) equals `on`

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,84 @@
...
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    return {
+      linkState: this.linkState,
+      checkedValue: true,
+      uncheckedValue: false
+    };
+  },
+  onChange: function (e) {
+    var val = e.target.checked ? this.props.checkedValue : this.props.uncheckedValue;
```
Pomax: `this.state. ...` rather than `this.props. ...`, since this value changes throughout the component's life cycle?

www_src/pages/editor/image-editor.jsx
k88hudson: React add `px` by default to many properties, including this one.

k88hudson: As for how to do this in a more reliable, consistent way, it requires a decision on the AST stuff we were talking about

www_src/pages/editor/text-editor.jsx
k88hudson: Again, requires a decision on the AST stuff etc that is outside the scope of this patch.

www_src/components/select/select.less
k88hudson: https://github.com/postcss/autoprefixer/issues/43

k88hudson: It will be added back in, in 5.2

Pomax: hm, I was more wondering about whether we could add it in ourselves in the mean time. Can't seem to find useful info on that going through the various docs there, though.

www_src/main.less
Pomax: are we using bootstrap? because that's a very bootstrap-specific thing, not generate browser behaviour?

www_src/components/option-panel/option-panel.jsx
Pomax: It sounds like these are labels, not state indicators, in which case we should rename these to `checkedLabel` and `uncheckedLabel` or something, to prevent confusion with the actual checked value that the checkbox yields.

www_src/pages/editor/image-editor.jsx
Pomax: oh neat! I didn't know that.

www_src/pages/editor/editor.jsx
k88hudson: I agree with you but can we do this when we integrate the real data stuff?

Pomax: yup

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,83 @@
...
+});
+
+var CheckboxSet = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
+  getDefaultProps: function () {
+    linkState: this.linkState
+  },
+  render: function () {
+    return (<OptionPanel>
+      {this.props.options.map((props, i) => <Checkbox {...props} key={i} linkState={this.props.linkState} />)}
```
Pomax: forgot to flag this earlier - I doubt it's extremely important since we're using pretty much "fixed lists" here, but you generally want a key that identifies the thing you're building, so in this cause that would be the option value, not its index (if the array is reordered but we use array index as key, React will diff everything, instead of going "these elements are simply reordered")

www_src/components/option-panel/option-panel.jsx
```diff
@@ -0,0 +1,83 @@
...
+  },
+  onChange: function (e) {
+    this.valueLink.requestChange(e.target.value);
+  },
+  render: function () {
+    this.valueLink = this.props.linkState(this.props.id);
+    var currentVal = this.valueLink.value;
+    return (<OptionPanel>
+      {this.props.options.map((item, i) => {
+        return (<label key={i} className={classNames('label', {selected: item.value === currentVal})}>
```
Pomax: same thing with the keys here

PR #1639, opened 2015-05-04
replacing hard coded color values with color variables

gvn:
There's still some hard coded values in the build, but I replaced the ones that had exact or close values in our palette.

### DIFF
--- a/www_src/pages/map/map.less
+++ b/www_src/pages/map/map.less
@@ -23,7 +23,7 @@
     transition: opacity 0.3s ease-in-out;
     width: 100%;
     height: 100%;
-    background: #F1D474;
+    background: @yellow;
     display: flex;
     align-items: center;
     justify-content: center;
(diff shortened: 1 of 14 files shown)


www_src/pages/map/map.less
```diff
@@ -32,20 +32,20 @@
...
-      box-shadow: 0 12px 10px -4px fade(darken(#E9EBF0, 30%), 50%);
+      box-shadow: 0 12px 10px -4px fade(darken(@lightGrey, 30%), 50%);
     }
     &.unselected {
       opacity: 0.5;
     }
     &.add {
-      background: #DCE0E6;
-      color: #C4C8CC;
+      background: darken(@lightGrey, 7%);
```
k88hudson: This looks too dark to me now

k88hudson:
Looks all good, except the add tiles -- those look a bit off to me

www_src/pages/map/map.less
gvn: What do you suggest?

k88hudson: `4%` is closer to the design i think

k88hudson:
Also, want to throw lightGrey in the background definition in main.less?

PR #1657, opened 2015-05-05
fixes single touch -> multi touch scale and rotation errors

### LINKED ISSUE #1635: Element scales & rotates unexpectedly
Steps to reproduce:
* Add text element to a page
* Move element to another spot on the page
* Without releasing the finger you used to move, place second finger anywhere on the screen
* Element scales & rotates unexpectedly

Hypothesis
* When second finger touches, it's using the original starting point of the element as it's reference point for the second touch, not the current location of the element

Pomax:
fixes #1635, which revealed single touch -> multi touch scale and rotation errors


thisandagain:
R+ :panda_face:

PR #1662, opened 2015-05-05
view fix for project

### LINKED ISSUE #1661: Page - Fails to load after adding an element
STR:
- Navigate to the page view
- Add any element
- Go back to the prior (project) view
- Return to the page view

Expected:
- Page view loads with element that was added and persisted to disk by the binding mixin

Action:
- Page view fails to load

Pomax:
fixes #1661


thisandagain:
R+ :goat:

PR #1663, opened 2015-05-05
relocated delete button to the button cluster

### LINKED ISSUE #1653: put the "Delete element" trash button in the right place
Due to some oddly hardcoded LESS the element delete button is currently in the upper left, but this should be moved to the button cluster at the bottom instead, next to the (+)

Pomax:
fixes #1653, relocates the delete button to the button cluster


gvn:
Looks good!

PR #1666, opened 2015-05-05
Requesting a pull to mozilla:v1 from k88hudson:data

k88hudson:


package.json
```diff
@@ -61,6 +61,8 @@
     "react": "0.13.1",
     "react-hammerjs": "0.2.2",
     "react-imageloader": "1.2.0",
+    "react-loadermixin": "1.0.4",
```
thisandagain: This can be removed ... looks like a conflicting change with #1662

k88hudson: ah, ok. showed up in my rebase conflict.

www_src/lib/api-fake-data.js
```diff
@@ -0,0 +1,58 @@
+module.exports = {
```
thisandagain: Mind if we call this `mock.js` or `mock.json` and keep it isolated to the project (map) page?

www_src/lib/api-mock.js
```diff
@@ -0,0 +1,73 @@
...
+      cb(null, {});
+    }
+  }
+
+  else {
+    cb(new Error('That uri does not exist'));
+  }
+
+  // Update Android
+  if (window.Android && ['put', 'post', 'delete'].indexOf(method) > -1) {
```
thisandagain: Good catch. I should add this to the actual API module as well. :+1:

k88hudson: maybe, but in the case of actual data, there's definitely a possibility that a GET response differs from the currently cached one. if that's what you mean.

thisandagain: I'm just thinking that we should *never* cache POST, PUT, or DELETE regardless of the `useCache` flag.

thisandagain:
Good start. R+ :poodle: :boom:

www_src/lib/api-mock.js
k88hudson: ah! right. good call

Issue #1645, opened 2015-05-04
Link editor


k88hudson:
The link editor still needs to be implemented on the page editor (which is currently `projects/projects.jsx`)

![image](https://cloud.githubusercontent.com/assets/1455535/7480048/b35e6102-f333-11e4-9074-342e4e879bc2.png)
(https://projects.invisionapp.com/share/BK2M18Q28#/screens/75741135)

Note that the actual link destination editor should be omitted, replaced with an alert (this is not done yet etc.)

xmatthewx:
Depends on #1549

k88hudson:
@xmatthewx is that issue not finalized?

xmatthewx:
@vazquez revised today, but I haven't had a chance to review the changes. This is listed as Beta 1/1 on roadmap. It's great if we can land this for Alpha, but I'm guessing the destination picker will require a bit of time.

xmatthewx:
...a bit of time to implement.

k88hudson:
Ah, ok. What is the plan for what to do for v1 then? Omit/hide the link element from the add menu etc.?
cc @thisandagain

thisandagain:
Let's show with alert (same as image selector). Hiding will make it seem "complete" which will likely raise a lot of questions and sets a lower expectation for the finished product.

xmatthewx:
This will include:
 - element editor for button style and label
 - destination picker

Build of tinker mode for destination will come later.

For alpha, it'd be nice to have the editor if it's not a major lift.

k88hudson:
Got it. I'll update the first comment.

k88hudson:
done

PR #1668, opened 2015-05-06
link editor

Pomax:
sets up the basics for fixing #1645, although it still needs to actually be hooked up to "do" things in conjunction with a real page.

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
+++ b/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
@@ -32,7 +32,7 @@ public class WebmakerApplication extends Application {
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
         Router.sharedRouter().map("/map/:projectId", Project.class);
         Router.sharedRouter().map("/projects/:projectId", Page.class);
-        Router.sharedRouter().map("/projects/:projectId/elements/:elementId", Element.class);
+        Router.sharedRouter().map("/elements/:projectId/:elementId/:type", Element.class);
         Router.sharedRouter().map("/projects/:projectId/elements/:elementId/:attributeId", Tinker.class);
 
         // @todo Restore state
--- a/www_src/lib/color.js
+++ b/www_src/lib/color.js
@@ -1,0 +1,17 @@
+var Color = require('color');
+var minimumContrast = 2;
+
+module.exports = {
+  getContrastingColor: (function() {
+    var white = Color().rgb(255, 255, 255),
+        whiteCSS = white.rgbString(),
+        black = Color().rgb(0, 0, 0),
+        blackCSS = black.rgbString();
+    return function(input) {
+      var c = Color(input),
+          wc = white.contrast(c),
+          bc = black.contrast(c);
+      return (wc < bc && wc >= minimumContrast) ? whiteCSS : blackCSS;
+    };
+  }())
+};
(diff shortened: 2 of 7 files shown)


app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
```diff
@@ -32,7 +32,8 @@ public void onCreate() {
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
         Router.sharedRouter().map("/map/:projectId", Map.class);
         Router.sharedRouter().map("/projects/:projectId", Project.class);
-        Router.sharedRouter().map("/projects/:projectId/elements/:elementId", Editor.class);
+        Router.sharedRouter().map("/edit/:projectId/:elementId/:type", Editor.class);
```
k88hudson: while i agree these routes need some work, probably doing in another ticket -- would you mind holding off? (will reduce merge conflicts too)

www_src/pages/editor/link-editor.jsx
```diff
@@ -0,0 +1,62 @@
...
+          <div className="form-group">
+            <label>Font</label>
+            <select className="select" valueLink={this.linkState('fontFamily')}>
+              <option value="Roboto">Roboto</option>
+              <option value="Bitter">Bitter</option>
+              <option value="Pacifico">Pacifico</option>
+            </select>
+          </div>
+          <div className="form-group">
+            <label>Color</label>
```
k88hudson: Designs say "Background Color"

www_src/pages/editor/link-editor.jsx
```diff
@@ -0,0 +1,62 @@
...
+var LinkEditor = React.createClass({
+  mixins: [
+    React.addons.LinkedStateMixin,
+    Binding
+  ],
+  getInitialState: function () {
+    return {
+      borderRadius: 5,
+      color: "#69A0FC",
+      textColor: "white",
```
k88hudson: I think what we actually need here is to compute whether white or black has greater contrast from the background color -- there are a bunch of libraries that can handle this such as https://www.npmjs.com/package/color. It should be pretty simple to do. would you mind adding that in?

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
Pomax: not really: the routing is necessary to have the correct element pull up the correct editor on Android, which cannot hook into window.location.has like editor.jsx does right now (That only works in the desktop browser)

www_src/pages/editor/link-editor.jsx
Pomax: agreed, will updated the code to make that happen automagically

www_src/pages/editor/link-editor.jsx
Pomax: whoops, good catch

www_src/lib/color.js
```diff
@@ -0,0 +1,17 @@
+var Color = require('color');
+var minimumContrast = 2;
+
+module.exports = {
+  getContrastingColor: (function() {
```
k88hudson: no need to reimplement .contrast https://www.npmjs.com/package/color#luminosity

Pomax:
resetting to "needs review" because I'm fairly certain the rebase went fine, but I'd like another R? pass just to be sure.

Issue #1649, opened 2015-05-04
Rename /project


k88hudson:
It should be /page or something.

thisandagain:
:+1: to renaming to `page` ... also `map` should be `project`.

k88hudson:
done

PR #1667, opened 2015-05-06
Expose router parameters to JS

thisandagain:

PR #1658, opened 2015-05-05
Update Android ActionBar styles

thisandagain:
This PR brings the navigation styles as close spec as possible without creating a completely custom ActionBar.

PR #1669, opened 2015-05-06
#1649 rename map to project, project to page, editor to element

k88hudson:


k88hudson:
@thisandagain names OK with you?

app/src/main/AndroidManifest.xml
```diff
@@ -23,37 +23,37 @@
...
             android:parentActivityName=".MainActivity"
             android:screenOrientation="portrait" >
         </activity>
         <activity
-            android:name=".activity.Project"
-            android:label="@string/title_activity_project"
-            android:parentActivityName=".activity.Map"
+            android:name=".activity.Page"
+            android:label="@string/title_activity_page"
+            android:parentActivityName=".activity.Project"
```
Pomax: activity.Page?

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
```diff
@@ -30,9 +30,9 @@ public void onCreate() {
         Router.sharedRouter().setContext(getApplicationContext());
         Router.sharedRouter().map("/main", MainActivity.class);
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
-        Router.sharedRouter().map("/map/:projectId", Map.class);
-        Router.sharedRouter().map("/projects/:projectId", Project.class);
-        Router.sharedRouter().map("/projects/:projectId/elements/:elementId", Editor.class);
+        Router.sharedRouter().map("/map/:projectId", Project.class);
```
Pomax: probably safer if we refactor this part now instead of later, so that Project.class lives on /projects, Page on /pages, etc

app/src/main/java/mozilla/org/webmaker/activity/Element.java
```diff
@@ -0,0 +1,10 @@
+package mozilla.org.webmaker.activity;
+
+import mozilla.org.webmaker.R;
+import mozilla.org.webmaker.WebmakerActivity;
+
+public class Element extends WebmakerActivity {
+    public Element() {
+        super("element", R.id.element_layout, R.layout.element_layout, R.menu.menu_element);
```
Pomax: awww yeah super classes

app/src/main/java/mozilla/org/webmaker/activity/Project.java
```diff
@@ -7,4 +9,16 @@
     public Project() {
         super("project", R.id.project_layout, R.layout.project_layout, R.menu.menu_project);
     }
+
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
```
Pomax: I think @thisandagain's patch will affect this, since it moves some onCreate things around. Worth looking at before we merge to see if there's conflicts or not there

app/src/main/res/values/strings.xml
```diff
@@ -3,9 +3,9 @@
 
     <string name="title_section1">Discover</string>
     <string name="title_section2">Make</string>
-    <string name="title_activity_map">Map</string>
     <string name="title_activity_project">Project</string>
-    <string name="title_activity_editor">Options</string>
+    <string name="title_activity_page">Page</string>
+    <string name="title_activity_element">Options</string>
```
Pomax: Do these two strings still make sense? (I got confused seeing the word options while editing my element, because I wasn't in an options menu, I was editing my content)

app/src/main/java/mozilla/org/webmaker/activity/Project.java
thisandagain: We should be ok here. Most of my changes to the `onCreateOptionsMenu` landed in #1658

app/src/main/res/values/strings.xml
thisandagain: I agree. This matches the spec more closely, but something to consider. I recommend using what Kate has here but filing a follow-up ticket.

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
thisandagain: :+1: I was hoping to land #1667 before we did this though as it's going to make my patch really difficult to merge.

k88hudson: Yep, no problem

thisandagain:
R+ :sheep: I'll file a follow-up to update the routes in `WebmakerApplication.java` after this and #1667 land.

thisandagain:
Follow-up ticket here: https://github.com/mozilla/webmaker-app/issues/1673

Issue #1652, opened 2015-05-04
Type - Embed Fira and update default font-family


thisandagain:
We are currently using Roboto. We need to embed Fira in the package and update the `main.less` file to use "Fira Sans".

k88hudson:
good call

k88hudson:
take a look at the master branch, all the code should still be there

k88hudson:
done

PR #1675, opened 2015-05-06
Use Fira as default typeface for app. Resolves GH-1652

thisandagain:


thisandagain:
@k88hudson R?

www_src/main.less
```diff
@@ -42,7 +42,7 @@ body {
   height: 100%;
   margin: 0;
   padding: 0;
-  font-family: 'Roboto', sans-serif;
+  font-family: "FiraSans", "Fira Sans", "Fira Sans OT", sans-serif;
```
k88hudson: what's the `"FiraSans"` for?

www_src/fonts.less
```diff
@@ -0,0 +1,21 @@
...
+
+@font-face {
+  font-family: 'Fira Sans';
+  src: local('Fira Sans Medium'),
+    url('@{fonts-path}/FiraSans-Medium.otf') format('otf'),
+    url('@{fonts-path}/FiraSans-Medium.woff') format('woff'),
+    url('@{fonts-path}/FiraSans-Medium.tff') format('truetype');
+  font-weight: 500;
+  font-style: normal;
+}
```
k88hudson: Where are the italics?

k88hudson:
I would add italics but otherwise this is good to go.

www_src/main.less
Pomax: why "FiraSans" (is that the name it has when system-installed)? Also for pure [a-zA-Z]+ strings, CSS doesn't need quote symbols. Ironically that means that "sans-serif" *would* needs quotes around it except it's a baked in value, and using quotes around it would entirely the wrong thing (fun times!)

www_src/fonts.less
```diff
@@ -0,0 +1,21 @@
+@fonts-path: '../fonts';
+
+@font-face {
+  font-family: 'Fira Sans';
+  src: local('Fira Sans Regular'),
+    url('@{fonts-path}/FiraSans-Regular.otf') format('otf'),
+    url('@{fonts-path}/FiraSans-Regular.woff') format('woff'),
```
Pomax: is this necessary for the platforms we're targeting? Every modern browser supports WOFF, there is no need anymore to fall back to "otf" and "ttf" versions "just in case" because that case won't arise. For android 4.2 and 4.3 WOFF is a problem, but then I would recommend falling back purely to the `otf` version. There is no need to also have `ttf` fallback.

Pomax:
it's worth addressing the italics and "the amount of data we need to bundle for fallbacks" by only specifying the font formats that we really need in this PR

www_src/fonts.less
thisandagain: Not needed according to @vazquez

www_src/main.less
thisandagain: Those are the three names used depending on which version of Fira is installed on the local system. This is a direct port from the `master` branch after we had various different problems with contributors and cross platform / browser issues with fonts.

www_src/fonts.less
thisandagain: Fair enough.

thisandagain:
As per @vazquez and our design specs we don't need italics.

Pomax:
R+ from me.

PR #1679, opened 2015-05-06
pinch zoom stuff

k88hudson:


www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
...
     var el = this.getDOMNode();
     var bounding = this.refs.bounding;
     var boundingEl = bounding.getDOMNode();
-    var startX, startY, deltaX, deltaY;
+    var scaleEl = this.refs.scaler.getDOMNode();
+    var isZoom, startX, startY, startDistance, deltaX, deltaY, currentZoom;
     var didMove = false;
+    var ZOOM_SENSITIVITY = 300;
 
     el.addEventListener('touchstart', (event) => {
```
Pomax: I'm a little worried about fragmentation - can we consolidate the code here with the code in `lib/touchhandler.js`?

www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
...
+        var x = this.state.camera.x;
+        var y = this.state.camera.y;
 
-      var translation = 'translate(' + (x + deltaX) + 'px, ' + (y + deltaY) + 'px)';
-      boundingEl.style.transform = translation;
+        deltaX = (event.touches[0].clientX - startX) / this.state.zoom;
+        deltaY = (event.touches[0].clientY - startY) / this.state.zoom;
+
+        var translation = 'translate(' + (x + deltaX) + 'px, ' + (y + deltaY) + 'px)';
+        boundingEl.style.transform = translation;
```
Pomax: There is no need to work with DOM nodes other than to get the intial bounds. Instead of setting the .style property manually, update the properties via setState() and and then let `render()` handle the diffing.

www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
...
+      }
 
     });
 
     el.addEventListener('touchend', (event) => {
       boundingEl.style.transition = '';
-      if (didMove) {
-        this.state.camera.x += deltaX;
-        this.state.camera.y += deltaY;
+      scaleEl.style.transition = '';
```
Pomax: same here. Reading from a DOM node = super useful. Changing things in a DOM node = rely on `setState` + `render`

www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
...
+    var isZoom, startX, startY, startDistance, deltaX, deltaY, currentZoom;
     var didMove = false;
+    var ZOOM_SENSITIVITY = 300;
 
     el.addEventListener('touchstart', (event) => {
       didMove = false;
       startX = event.touches[0].clientX;
       startY = event.touches[0].clientY;
       boundingEl.style.transition = 'none';
+      scaleEl.style.transition = 'none';
```
Pomax: There is no need to work with DOM nodes other than to get the intial bounds. Instead of setting the .style property manually, update the properties via setState() and and then let `render()` handle the diffing.

www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
...
+        var distance = Math.sqrt(dx*dx + dy*dy);
 
-      var x = parseInt(this.state.camera.x);
-      var y = parseInt(this.state.camera.y);
+        currentZoom = currentZoom + ((distance - startDistance) / ZOOM_SENSITIVITY);
+        currentZoom = Math.min(Math.max(currentZoom, 0.25), 1);
 
-      deltaX = (event.touches[0].clientX - startX) / this.state.zoom;
-      deltaY = (event.touches[0].clientY - startY) / this.state.zoom;
+        scaleEl.style.transform = 'scale(' + currentZoom + ')';
```
Pomax: ^

www_src/pages/project/project.jsx
```diff
@@ -43,39 +43,74 @@ var Map = React.createClass({
       });
     });
   },
+
   componentDidMount: function () {
     var el = this.getDOMNode();
     var bounding = this.refs.bounding;
     var boundingEl = bounding.getDOMNode();
```
Pomax: looking at what we do with this var, as well as `scaleEl`, rewriting this to using setState+render to take care of the correct styling as you're interacting (see the `Positionable.jsx` component) means we don't need to even extract these two DOM nodes.

www_src/pages/project/project.jsx
k88hudson: Performance is way better if you bypass the state machine for every single touch update, and only update state after the interaction is finished. This especially true as the render function increases in complexity, which it is in this view.

In fact, I would recommend we refactor the code in the page editor to follow this pattern,

www_src/pages/project/project.jsx
k88hudson: next heartbeat, maybe -- refactoring to use multipurpose touch handlers is not easy

thisandagain:
This is awesome. Aside from the origin piece, this feels :+1: :ok_hand: 

One detail that is a little off from @flukeout 's spec is that the user can currently zoom too far in which makes distinguishing it from the page view confusing. You can see the max zoom-in state in the latest diagram: https://dl.dropboxusercontent.com/u/109902/tiles/zoom-ux-3.pdf

thisandagain:
R+ if you'd like to land and work on those two details separately. :panda_face:

thisandagain:
One more detail I just found ... tapping again on a selected page deselects all pages. In spec, I think this should do nothing.

PR #1681, opened 2015-05-06
removes a nullpointer exception when retriggering oncreate without extras

Pomax:
we missed a potential null pointer, so going from project to page and then back will crash webmaker-app =x


thisandagain:
Nice catch! :smile:

PR #1682, opened 2015-05-07
make it possible to edit text with a tap

### LINKED ISSUE #1654: Interim text editor
We need something to edit the text content in the editor view.

Let's use an edit dialog for now.

Pomax:
fixes #1654 by switching over to the WebChromeClient, which lets us use alert, prompt, and confirm (based on [this SO question](https://stackoverflow.com/questions/5271898/javascript-alert-not-working-in-android-webview)).

This probably works well enough for alpha, but will need a followup to do "real dialogs" and a revert to WebView since this solution relies on (a) Android 5 and (b) having Chrome installed. Reasonable for alpha, not so reasonable for after that.

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/view/WebmakerWebView.java
+++ b/app/src/main/java/mozilla/org/webmaker/view/WebmakerWebView.java
@@ -5,6 +5,9 @@ import android.content.Context;
 import android.view.ViewGroup;
 import android.webkit.WebView;
 
+import android.webkit.JsPromptResult;
+import android.webkit.WebChromeClient;
+
 import mozilla.org.webmaker.client.WebClient;
 import mozilla.org.webmaker.javascript.WebAppInterface;
 import org.json.JSONObject;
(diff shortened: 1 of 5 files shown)


app/src/main/java/mozilla/org/webmaker/view/WebmakerWebView.java
```diff
@@ -5,6 +5,9 @@
 import android.view.ViewGroup;
 import android.webkit.WebView;
 
+import android.webkit.JsPromptResult;
+import android.webkit.WebChromeClient;
```
Pomax: this probably requires the same extensions that mozilla.org.webmaker.client.WebClient got, come to think of it...

www_src/pages/page/blocks/Text.jsx
```diff
@@ -11,6 +11,18 @@ var Header = React.createClass({
 
   render: function() {
     return <p>{this.state.value}</p>;
+  },
+
+  handleTap: function(diff) {
+    console.log("tap tap tap "+diff);
```
Pomax: rogue console log O_O

k88hudson:
Thanks for figuring this out

Issue #1689, opened 2015-05-07
On back, route params are empty


k88hudson:
After pressing the back button I don't seem to be actually getting any route params being returned, just an empty object

k88hudson:
Hm, actually i think i know why...

k88hudson:
Ok, this seems to only be an issue when you press the back button.

k88hudson:
@thisandagain any idea why this might be the case?

k88hudson:
Definitely related to this line being `null`
https://github.com/mozilla/webmaker-app/blob/v1/app/src/main/java/mozilla/org/webmaker/WebmakerActivity.java#L36

thisandagain:
The issue here is that the back button doesn't trigger the router and thus the route params are never attached to the intent extras (thus `intentExtras` = `null`).

thisandagain:
As per Vidyo call: will resolve by persisting / restoring route params to `SharedPreferences` via react mixin.

k88hudson:
Hurray! fixed!

PR #1690, opened 2015-05-07
Requesting a pull to mozilla:v1 from k88hudson:datapageeditor

k88hudson:
* Hooks up the page editor to persist data, read params
* Passes correct element/page id to the element editor

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
+++ b/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
@@ -32,8 +32,8 @@ public class WebmakerApplication extends Application {
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
         Router.sharedRouter().map("/projects/:project", Project.class);
         Router.sharedRouter().map("/projects/:project/pages/:page", Page.class);
-        Router.sharedRouter().map("/projects/:project/pages/:page/elements/:element/editor/:type", Element.class);
-        Router.sharedRouter().map("/projects/:project/pages/:page/elements/:element/attributes/:attribute/editor/:type", Tinker.class);
+        Router.sharedRouter().map("/projects/:project/pages/:page/elements/:element/editor/:editor", Element.class);
+        Router.sharedRouter().map("/projects/:project/pages/:page/elements/:element/attributes/:attribute/editor/:editor", Tinker.class);
 
         // @todo Restore state
     }
--- a/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
+++ b/app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
@@ -30,10 +30,10 @@ public class WebmakerApplication extends Application {
         Router.sharedRouter().setContext(getApplicationContext());
         Router.sharedRouter().map("/main", MainActivity.class);
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
-        Router.sharedRouter().map("/map/:projectId", Project.class);
-        Router.sharedRouter().map("/projects/:projectId", Page.class);
-        Router.sharedRouter().map("/elements/:projectId/:elementId/:type", Element.class);
-        Router.sharedRouter().map("/projects/:projectId/elements/:elementId/:attributeId", Tinker.class);
+        Router.sharedRouter().map("/projects/:project", Project.class);
+        Router.sharedRouter().map("/projects/:project/pages/:page", Page.class);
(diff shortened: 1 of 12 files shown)


k88hudson:
Doesn't seem to be working because of https://github.com/mozilla/webmaker-app/issues/1689

k88hudson:
Actually still need to fix a couple of things

k88hudson:
@thisandagain Ok, the only remaining issue with page editor persistence seems to be router state doesn't get restored on back.

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -29,6 +29,9 @@ public WebAppInterface(Context c, JSONObject routeParams) {
         mPrefs = mContext.getSharedPreferences(mPrefKey, 0);
         mPageState = mPrefs.getString("page_state", "{}");
         mRoute = routeParams;
+        if (routeParams != null) {
```
thisandagain: Probably better to log this in `WebmakerActivity` where the object is first created instead.

thisandagain:
:beetle: One bug that I'm noticing is that after adding elements to a page, if you exit, go to another page, and then return those elements are rendered at a different origin / position.

STR:
- Enter project view
- Open page at 0,0
- Add three elements (rotate and scale them)
- Exit page and open page at 0,1
- Add one element (don't rotate and scale)
- Exit and then return to both pages

Expected:
- Elements are in the position you left them in

Actual
- Elements seem to be shifted into the top left corner

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
k88hudson: Ah, sorry. this is leftover -- i'll remove

k88hudson:
Huh. that's totally weird

k88hudson:
Looks like for some reason the page editor is adding `transform-origin: center 50% 0px` to elements when they are moved, but this property isn't being persisted to state. I'm trying to figure out where the heck this might get added...

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
```diff
@@ -30,10 +30,10 @@ public void onCreate() {
         Router.sharedRouter().setContext(getApplicationContext());
         Router.sharedRouter().map("/main", MainActivity.class);
         Router.sharedRouter().map("/main/:tab", MainActivity.class);
-        Router.sharedRouter().map("/map/:projectId", Project.class);
-        Router.sharedRouter().map("/projects/:projectId", Page.class);
-        Router.sharedRouter().map("/elements/:projectId/:elementId/:type", Element.class);
-        Router.sharedRouter().map("/projects/:projectId/elements/:elementId/:attributeId", Tinker.class);
+        Router.sharedRouter().map("/projects/:project", Project.class);
+        Router.sharedRouter().map("/projects/:project/pages/:page", Page.class);
```
Pomax: I'd recommend making these match the thing they're loading instead of wildly overloading the /projects route and using a mixture of :params and virtual directory labels to hope things work out.

k88hudson:
ok, it looks like the `positionableContainer` is not restoring the right left/top values.

e.g. they should be: `left: 150px; top: 198.5px;` but are actually `left: -10px; top: -21.5px`

k88hudson:
Ok, looks like the reason is that the `Positionable` component has some state that doesn't get persisted to the actual element  (`xoffset` and `yoffset`) so that will need some refactoring

k88hudson:
@thisandagain Ok, new router stuff integrated

app/src/main/java/mozilla/org/webmaker/WebmakerApplication.java
thisandagain: Not sure I 100% understand, but these match the API routes and we need each of these params to make API requests. Do you have an example of how you would structure this differently?

k88hudson: i'm just doing this in the style of the API, but feel free to file this as a separate issue on this/the API

thisandagain:
Getting an exception when attempting to enter tinker mode:
```bash
[INFO:CONSOLE(306)] "Uncaught Error: Java exception was raised during method invocation", source: file:///android_asset/www/js/element.bundle.js (306)
```

thisandagain:
R+ !!!! :ok_hand:

PR #1693, opened 2015-05-07
Use the right file names.

### LINKED ISSUE #1692: inconsistent casing between filenames and requires()
e.g.in www_src/pages/page/page.jsx we have var Positionable = require('./positionable.jsx');
but the actual name of the file is Positionable.jsx.
This is the problem for those of us who use case-sensitive OS's like Linux!
Loading the app throws errors about 'positionable.jsx not found' because of it.

errietta:
Closes #1692


thisandagain:
I feel like the right fix for this is actually to change the names of the files themselves rather than to update the `require` statements. Lowercase filenames for react components is more in keeping with our style conventions.

thisandagain:
R+ :panda_face: :cake: :cat2:

PR #1694, opened 2015-05-07
Integrate router into react state. Resolves GH-1689

thisandagain:

### DIFF
--- a/www_src/lib/router.jsx
+++ b/www_src/lib/router.jsx
@@ -10,13 +10,10 @@ module.exports = {
 
       // If they do, save them. If not, fetch from SharedPreferences
       if (routeParams !== '{}') {
-        console.log('***** Route params exist *****');
         r = JSON.parse(routeParams);
         this.android.setSharedPreferences(key, routeParams, true);
       } else {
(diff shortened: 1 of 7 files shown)


thisandagain:
@k88hudson R?

www_src/lib/router.jsx
```diff
@@ -0,0 +1,25 @@
...
+      // Check to see if route params exist & create cache key
+      var routeParams = this.android.getRouteParams();
+      var key = 'state::route';
+
+      // If they do, save them. If not, fetch from SharedPreferences
+      if (routeParams !== '{}') {
+        r = JSON.parse(routeParams);
+        this.android.setSharedPreferences(key, routeParams, true);
+      } else {
+        var hit = this.android.getSharedPreferences(key, true);
```
k88hudson: i cant remember, is it possible this could be `null`?

k88hudson:
Ok, this looks good to me

www_src/lib/router.jsx
thisandagain: I couldn't think of how, but probably best to be safe than sorry.

PR #1704, opened 2015-05-08
cache and restore page state correctly

### LINKED ISSUE #1696: Positionable - Cannot restore state
First noticed in #1690 

:beetle: One bug that I'm noticing is that after adding elements to a page, if you exit, go to another page, and then return those elements are rendered at a different origin / position.

STR:
- Enter project view
- Open page at 0,0
- Add three elements (rotate and scale them)
- Exit page and open page at 0,1
- Add one element (don't rotate and scale)
- Exit and then return to both pages

Expected:
- Elements are in the position you left them in

Actual
- Elements seem to be shifted into the top left corner

Pomax:
fixes #1696

### DIFF
--- a/www_src/pages/page/page.jsx
+++ b/www_src/pages/page/page.jsx
@@ -17,26 +17,26 @@ var Project = React.createClass({
     return {
       content: [],
       currentElement: -1,
-      showAddMenu: false
+      showAddMenu: false,
+      dims: {
+        width: 0,
+        height: 0
+      }
     };
   },
 
   componentWillMount: function() {
     if (this.state.params.page) {
       api({
+        method: 'get',
         uri: '/users/foo/projects/bar/pages/' + this.state.params.page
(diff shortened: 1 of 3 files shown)


www_src/pages/page/page.jsx
```diff
@@ -17,26 +17,26 @@ var Project = React.createClass({
...
     }
-
-    this.dims = {
-      width: 0,
-      height: 0
-    };
   },
 
-  componentDidUpdate: function () {
+  componentDidUpdate: function (prevProps, prevState) {
```
k88hudson: i don't think you need this change

k88hudson:
Ok looks good! Thanks for fixing this

www_src/pages/page/page.jsx
Pomax: away they go.

Issue #1695, opened 2015-05-08
Android - Soft and hard back buttons behave differently


thisandagain:
The "hard" Android back button performs as you would expect by dismissing the active Activity and going back to the parent in the stack. Currently, the "soft" back button seems to be re-triggering the Activity life-cycle which is slower and has possible memory allocation repercussions.

Needs more investigation.

PR #1708, opened 2015-05-08
Persist element definitions

k88hudson:
This is working for images and text so far.


www_src/blocks/generator.js
```diff
@@ -11,18 +12,19 @@ Generator.IMAGE = "image";
 
 Generator.generateBlock = function(options) {
```
k88hudson: Personally I would rather get rid of this function altogether and just export the block components

k88hudson: Actually eyah i think i'm gonna do that

package.json
```diff
@@ -62,7 +62,7 @@
     "react": "0.13.1",
     "react-hammerjs": "0.2.2",
     "react-imageloader": "1.2.0",
-    "url-pattern": "^0.8.0",
+    "url-pattern": "git://github.com/k88hudson/url-pattern.git#hyphenunderscore",
```
Pomax: what does this change and is there already something that does this so we don't have to point to personal git repos?

www_src/blocks/generator.js
```diff
@@ -2,6 +2,7 @@ var React = require('react');
 var Text = require('./text.jsx');
 var Link = require('./link.jsx');
 var Image = require('./image.jsx');
+var uuid = require('../lib/uuid');
```
Pomax: why not https://www.npmjs.com/package/node-uuid?

www_src/blocks/generator.js
```diff
@@ -11,18 +12,19 @@ Generator.IMAGE = "image";
...
-    return <Text value={options.value} />;
+    return <Text {...options} />;
   }
   if (options.type === Generator.LINK) {
-   return <Link href={options.href} label={options.label} active={options.active} />;
+   return <Link {...options} />;
   }
   if (options.type === Generator.IMAGE) {
-    return <Image src={options.src} alt={options.alt} />;
+    return <Image {...options} />;
```
Pomax: Does ES6 +JSX give us some way to compact this to something like `<${type} {...options}/>`? =/

www_src/blocks/image.jsx
```diff
@@ -0,0 +1,29 @@
...
+    defaults: {
+      src: '',
+      alt: '',
+      opacity: 1,
+      borderWidth: 0,
+      borderColor: 'transparent'
+    }
+  },
+  getDefaultProps: function () {
+    return this.defaults;
```
Pomax: since it's statics, `return Image.defaults`?

www_src/blocks/image.jsx
```diff
@@ -0,0 +1,29 @@
...
+  },
+  render: function() {
+    var style = {
+      opacity: this.props.opacity,
+      borderStyle: 'solid',
+      borderWidth: this.props.borderWidth,
+      borderColor: this.props.borderColor
+    };
+    return <img style={style} src={this.props.src} alt={this.props.alt} />
```
Pomax: do we want to use the slightly more complex LoaderImage that @thisandagain created?

www_src/blocks/text.jsx
```diff
@@ -0,0 +1,28 @@
...
+      fontWeight: 'normal',
+      fontStyle: 'normal',
+      textDecoration: 'none',
+      textAlign: 'center',
+      innerHTML: 'Hello world'
+    }
+  },
+
+  getDefaultProps: function () {
+    return this.defaults;
```
Pomax: see above

www_src/blocks/text.jsx
```diff
@@ -0,0 +1,28 @@
...
+  },
+
+  render: function() {
+    var style = {};
+    ['fontSize', 'fontFamily', 'color', 'fontWeight', 'fontStyle', 'textDecoration', 'textAlign'].forEach(prop => style[prop] = this.props[prop]);
```
Pomax: `Object.keys(Text.defaults)`.forEach? To keep things DRY?

www_src/blocks/generator.js
k88hudson: we already have this in the repo

package.json
k88hudson: I had to land some changes, hopefully they'll be merged by next week

www_src/blocks/image.jsx
k88hudson: not in this patch preferably

www_src/blocks/image.jsx
k88hudson: `Image` is undefined at this point, also it appears statics get attached to instances as well, so what's the downside?

www_src/blocks/text.jsx
k88hudson: the problem is `innerHTML` is not included here. once we refactor to have style and attributes in separate objects this should be a lot cleaner.

www_src/components/range/range.jsx
```diff
@@ -18,10 +19,11 @@ var Range = React.createClass({
   render: function () {
     var linkState = this.props.linkState || this.linkState;
     var valueLink = linkState(this.props.id);
+
     return (
       <div className="range">
-        <input min={this.props.min} max={this.props.max} type="range" valueLink={valueLink}/>
-        <div className={'range-summary' + (parseInt(valueLink.value, 0) === this.props.min ? ' min' : '')}>{valueLink.value}{this.props.unit}</div>
+        <input min={this.props.min} max={this.props.max} step={this.props.step} type="range" valueLink={valueLink}/>
+        <div className={'range-summary' + (parseFloat(valueLink.value, 10) === this.props.min ? ' min' : '')}>{valueLink.value}{this.props.unit}</div>
```
k88hudson: good call

package.json
k88hudson: if you can think of a better alternative i'm all for it

www_src/blocks/image.jsx
thisandagain: We can add it in a follow-up. The loading state still hasn't been spec'd.

thisandagain:
This is working fairly well for me as long as I use the "soft" back button which triggers a new `onLoad` event for the parent activity. If you use the Android "hard" back button the prior view is shown without the changes being reflected. Related to #1695

package.json
Pomax: just worried we never remember to switch back. Do we have a tracking issue for this?

www_src/blocks/image.jsx
Pomax: when this function gets called, Image will be defined.

Issue #1722, opened 2015-05-09
Apply position to elements in /project


k88hudson:
Right now they aren't being transformed/scaled/rotated appropriately

k88hudson:
Dooone

PR #1723, opened 2015-05-09
[#1722] Add positioning to elements in /project view

k88hudson:


www_src/lib/propUtils.js
```diff
@@ -0,0 +1,18 @@
+var utils = {};
+
+utils.propsToPosition = function (props) {
```
thisandagain: :+1:

thisandagain:
One major issue I'm finding is that button elements in the project (map) view are still interactive. Taping them will take you to a 404. Otherwise, I think this is a good solution in place of having the thumbnails. R+ :smile:

Issue #1725, opened 2015-05-09
Alpha - Remove tinker mode button


thisandagain:
Remove tinker mode buttons from the alpha release. Restore once color components are functional.

PR #1730, opened 2015-05-09
Remove tinker mode button from color-group component. Resolves GH-1725

thisandagain:


www_src/components/color-group/color-group.jsx
```diff
@@ -38,11 +38,6 @@ var ColorGroup = React.createClass({
           <input className="sr-only" name="color" type="radio" value={color} checked={this.valueLink.value === color ? true : null} onChange={this.onChange} />
         </label>
       })}
-      <div className="tinker-container">
```
k88hudson: actually, what do you think about just hiding this instead? like with style display none or something. then we dont have to dig through the commit logs to reimplement

thisandagain: Sure. Not a problem.

PR #1734, opened 2015-05-09
Add ability to edit link text and refactor alert

k88hudson:


thisandagain:
That works! I like the alert refactor ... that makes more sense. R+

Issue #1824, opened 2015-05-12
Remove yellow adjacent page edges from the Page Editor view.


flukeout:
Since the edges currently don't represent what tiles are actually adjacent to the page being edited, and you can't navigate to other pages in this view yet, let's remove them altogether because they're confusing.

thisandagain:
:+1:

PR #1827, opened 2015-05-12
Hide adjacent pages from page view pending swipe implementation. 

thisandagain:
Resolves GH-1824


flukeout:
R+ The page edges are indeed hidden.

Issue #1815, opened 2015-05-11
Font Size attribute for Text Element seems pointless since I can pinch-scale


flukeout:
Since for now we're not using a 'bounding box' approach to text elements, there doesn't really seem to be a need for the Font Size setting.

xmatthewx:
Slider should at least reflect changes made with pinch transform. 

Until we're ready to implement bounding box, 
 - we can hide font-size setting
 - we should alignment

thisandagain:
Agreed that we should remove the font-size slider. Confirmed by quite a few test users today as well (via backchannel).

xmatthewx:
> we should aligment

Ack. We also should removed text alignment options

PR #1828, opened 2015-05-12
Remove font-size range control from text element editor.

thisandagain:
Resolves GH-1815


flukeout:
R+ Font Size control is gone.

Issue #1699, opened 2015-05-08
Better selected page state UI


k88hudson:
Can we get something more prominent than this? The shadow doesn't stand out enough when the grid is filled.

![image](https://cloud.githubusercontent.com/assets/1455535/7540738/eb45f9f8-f57d-11e4-8d34-510195dd67db.png)

xmatthewx:
In the comp we see shadow on 3 sides. In the build it's more transparent and only below the page. Pages also feel closer in the build by a few pixels, a bit crowded.

vazquez:
@flukeout @xmatthewx - What do you think of this?

![tile map view tile selected](https://cloud.githubusercontent.com/assets/1119821/7574272/1208d56e-f7fa-11e4-8160-389bcc230bda.png)

flukeout:
I like this treatment!
Is it worth mocking it up against a darker background to future-proof it?

vazquez:
Perfect, will do

vazquez:
@flukeout What do you think? I think we're still good :)

![screenshot 2015-05-11 16 36 02](https://cloud.githubusercontent.com/assets/1119821/7574521/eb552902-f7fb-11e4-8e33-8ce1d0a21780.png)

thisandagain:
WFM. Big improvement over existing selection indicator. :+1:

PR #1829, opened 2015-05-12
Update selected page style.

thisandagain:
Resolves GH-1699

### DIFF
--- a/www_src/pages/project/project.less
+++ b/www_src/pages/project/project.less
@@ -26,8 +26,10 @@
     position: absolute;
     left: 0;
     top: 0;
+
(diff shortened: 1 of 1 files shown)


flukeout:
@thisandagain Good to merge, but minor nitpick is that the selected page shrinks a bit when it receives the border. I think @vazquez intended for the page to stay the same size. If it's just a matter of the border rule making this happen, one idea is to add a transparent border to the non-selected tiles and then just changing the color.

thisandagain:
I tried that and it absolutely killed performance for some reason. I can investigate further in a follow-up ticket.

Issue #1822, opened 2015-05-11
Allow canceling out of the Add Element view by tapping somewhere away from the options.


flukeout:
I think some of our other UX (like selecting image source) will allow this behaviour, so we should let users cancel adding an element by either tapping the (X) icon on the bottom, or anywhere on the dark overlay.

thisandagain:
:+1: :+1: :+1:

PR #1830, opened 2015-05-12
Allow tapping overlay to dismiss add element mode.

thisandagain:
Resolves GH-1822


flukeout:
@thisandagain Works as expected. One thing that is a bit weird is that the back arrow in the header returns you to the Map view from the Add Element view. That might be a bit confusing, but I'll file a separate feedback issue.

thisandagain:
Good call. Mind adding that detail to the back button UX ticket? I'll try to hit that this sprint as well.

PR #1831, opened 2015-05-12
[WIP] seamless text input in editor view

Pomax:
do not merge

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "react": "0.13.1",
     "react-hammerjs": "0.2.2",
     "react-imageloader": "1.2.0",
+    "react-onclickoutside": "^0.2.4",
     "url-pattern": "git://github.com/k88hudson/url-pattern.git#hyphenunderscore",
     "xhr": "2.0.1"
   }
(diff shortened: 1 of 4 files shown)


Pomax:
This will claim things like:

```
05-12 12:36:51.110    1627-1627/mozilla.org.webmaker I/chromium﹕ [INFO:CONSOLE(893)] "saving {"id":"bartext","innerHTML":"such Hello world","type":"text","angle":0,"parentHeight":440,"parentWidth":320,"scale":1,"x":50,"xoffset":-55.1171875,"y":350,"yoffset":-20.5,"zIndex":2,"opacity":0.5,"ref":"positionable2","key":"positionable2","current":true,"interactive":true,"touchactive":true,"fontSize":18,"fontFamily":"Roboto","color":"#645839","fontWeight":"normal","fontStyle":"normal","textDecoration":"none","textAlign":"center","editing":false}", source: file:///android_asset/www/js/element.bundle.js (893)
```

which clearly has the right `innerHTML` value, but then when you navigate back to the Page view, the element still uses the old text.

PR #1839, opened 2015-05-12
improved text editing for text elements

Pomax:
Due to how Android works with its views, this can lead to a content flash when the page is restored after editing and the `<Text>` element gets updated, but only after it's been restored with the original text.

This also needs a bit of a CSS tweek on the page itself, because neither the page preview on "project" nor the element preview in the element editor have any text wrapping, whereas the page text element does, for some reason.

Testing:

- go to a project
- go to a page in that project
- add a text element to the page if there isn't one already
- select the text element and tap the edit button
- in the editor view, tap the text to now get "native android" text input

Missing due to updated ticket: "edit text" button, and the "minimum text length 1" restriction

### DIFF
--- a/www_src/blocks/text.jsx
+++ b/www_src/blocks/text.jsx
@@ -30,10 +30,8 @@ var Text = React.createClass({
   render: function() {
     var style = {};
     var props = this.props;
-    Object.keys(props).forEach(prop => {
-      if (prop === "innerHTML") return;
-      style[prop] = props[prop]
-    });
+    ['fontFamily', 'color', 'fontWeight', 'fontStyle', 'textDecoration', 'textAlign', 'whiteSpace']
+      .forEach(prop => style[prop] = props[prop]);
 
     if (props.position) {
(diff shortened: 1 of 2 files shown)


flukeout:
@Pomax questions & observations:
* When I tap the text in the preview to start editing, it shifts over to the right a bit - should remain in the same place as before
* Is there a way to add a "Done" action on the native keyboard

Unrelated to your work, but they grey UI of the Edit Element page blends in with the native keyboard. I'll address that with Ricardo.

flukeout:
Added #1844 to discuss the keyboard UI issue.

Pomax:
1. I can't reproduce that... maybe we can investigate on the nexus 4 tomorrow?
2. no, tapping outside the element is the general "okay I am done" in the absence of an explicit done button

flukeout:
I think we need a clearer action for finishing editing then. I've added that to #1840 - we'll have it figured out tomorrow.

thisandagain:
Really nice, clean solution. R+ :panda_face:

PR #1844, opened 2015-05-13
Google Analytics Implementation

### LINKED ISSUE #1472: v1 - Re-Integrate Universal Analytics
https://developers.google.com/analytics/devguides/collection/android/v4/

/cc @adamlofting

ryanw-se:
Has been thoroughly tested by myself locally, could use some more testing online however. The app name and id are specified in the strings.xml file, because they need to be manually set, otherwise it defaults to the app_name resource instead.

Google Analytics creates a new tracker from the app_tracker.xml file located in the ```res/xml``` folder. This folder contains the tracking id and screen names for each of the activities provided. Screen names have been set for already existing activities, for easier reading. 

This will need to be done for any future activities that are implemented, to stick to the convention. All Google Analytics implementation related stuff is stored in the ```WebmakerApplication.java``` class. For ease of manipulation later on, methods have been provided for both instances, the tracker and GA.

Dry run has been set to false, although its highly recommended that it is set to true while debugging. This will fire the events locally and not send them directly to the server. Good for debugging purposes and seeing how the data is sent before its actually sent to the server.

Closes issue: #1472 @thisandagain

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,7 +27,7 @@ android {
 
 dependencies {
     compile fileTree(dir: 'libs', include: ['*.jar'])
-    compile 'com.android.support:support-v4:22.1.1'
+    compile 'com.android.support:support-v13:22.1.1'
     compile 'com.android.support:appcompat-v7:22.1.1'
     compile 'com.google.android.gms:play-services-analytics:7.3.0'
 }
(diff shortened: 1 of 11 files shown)


thisandagain:
Confirmed data is coming through in Google Analytics. Ready to land. R+ :cat: :cake: :boom: :sheep:

thisandagain:
Thanks x1000 @ryanw-se

adamlofting:
I'd like to multiply that thanks by another x 1000 :smile: this is awesome.

 :sparkling_heart:

PR #1845, opened 2015-05-13
speed up webpack

k88hudson:
Because hurray for faster builds

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,4 +1,5 @@
 var getPages = require('./npm_tasks/get-pages');
+var path = require('path');
 
 // Prep all entry points
 var entry = {};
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -17,15 +17,18 @@ module.exports = {
...
         test: /\.jsx$/,
-        loaders:  ['babel-loader', 'jsx-loader']
+        loaders:  ['babel-loader', 'jsx-loader'],
+        include: __dirname + '/www_src'
       },
       {
         test: /\.json$/,
-        loader: 'json-loader'
+        loader: 'json-loader',
+        include: [__dirname + '/www_src',  __dirname + '/node_modules/color']
```
Pomax: I recommend against any "doing things to files inside `node_modules`".

k88hudson: unfortunately there is no way around this since colors requires json

k88hudson: what do you mean by "run from cache"?

Pomax: perhaps I misunderstand what's going on here, then - what does this step do that requires it to reach into the dependency directory? (because that dir should for pretty much all non-require purposes be a black box)

Pomax:
rawr+

PR #1846, opened 2015-05-13
Edit button

Pomax:
- edit button that toggles input mode
- fixed stylinng for smaller displays like the nexus 4


flukeout:
@Pomax Looks good - a few points:
* When you're editing the text and tap away from it, the keyboard goes away but the action button still says "Done" - can we officially end editing then and make the button say "Edit Text" again?
* Capitalize "Done" and "Edit Text" labels on the action button
* Down the road, we'll want to add some icons to the buttons
* Can we keep the behaviour of tapping the text to initiate editing in addition to the "Edit Text" button? Now that it's gone, I miss it

Pomax:
1. fixed
2. fixed
3. agreed. do we have an issue for this already?
4. hehe... added back in =)

thisandagain:
:+1: Feeling better! Thanks folks.

PR #1869, opened 2015-05-14
API Integration

k88hudson:


k88hudson:
Alright so this is what we've got

### General flow
* except for the discover view and tinker, all the views have GET, POST, PATCH, and DELETE api calls hooked up.
* when a view resumes or is created, it synchronously waits for the appropriate data to be returned 
* in most cases, api updates are performed at the time they are triggered (creating/deleting pages or elements, etc.)
* for elements on the page view, updates are performed on touchEnd events
* for the element editor, updates are performed when the view is paused

### Element spec
* since react requires a flat structure for components in order to take advantage of defaulting, merging state updates, etc. there is a data transformation layer between components that have two way binding
* data that is *received * from the API is "flattened"
* data that is *sent* to the API is "expanded" into style and attribute properties
* the formatting, defaults, and names of accepted values for each element are validated according to a "spec", which is defined in each block component 
* all this functionality is included in lib/spec.js, with tests in tests/spec.test.js

### Page refactor
* in had to decouple coordinates for the Positionable component from the notion of being centred on the page, reducing dims/x-offset/y-offset to just x and y coordinates.

TODO:
* loading states, error states when data is not found
* error state or alternative view if you go offline
* centre elements when you first add them.

www_src/blocks/all.jsx
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
k88hudson: Just a utility to get the right block, since this is used a lot. Replaces `generator.js` below

www_src/blocks/image.jsx
```diff
@@ -1,20 +1,45 @@
...
-var Image = React.createClass({
-  statics: {
-    defaults: {
-      src: '',
-      alt: '',
-      opacity: 1,
-      borderWidth: 0,
-      borderColor: 'transparent'
-    }
+var spec = new Spec('image', assign({
```
k88hudson: This is just a set of allowed properties, including defaults, which allows for transformation between the API format and the react format -- I was super worried about doing transformations without having validations and tests on this

www_src/blocks/link.jsx
```diff
@@ -1,24 +1,50 @@
...
-  statics: {
-    defaults: {
-      href: '',
-      borderRadius: 5,
-      backgroundColor: '#69A0FC',
-      fontFamily: 'Roboto',
-      innerHTML: 'Tap me',
-      active: false
-    }
+var spec = new Spec('link', assign({
```
k88hudson: Same thing here. just defining allowed props.

www_src/components/range/range.jsx
```diff
@@ -16,13 +16,16 @@ var Range = React.createClass({
       value: 100
     };
   },
+  onChange: function (e) {
+    this.valueLink.requestChange(parseFloat(e.target.value));
```
k88hudson: This change allows us to always return an actual number from ranges, instead of a string -- necessary for prop type validation

www_src/lib/api.js
```diff
@@ -1,22 +1,23 @@
...
     useCache: false,
     json: {},
-    timeout: 60000    // 60 seconds
+    headers: {},
+    timeout: 60000 // 60 seconds
   });
 
   // ensure user-supplied methods conform to what we need.
-  options.method = options.method.toLowerCase();
+  options.method = options.method.toUpperCase();
```
k88hudson: xhr does not normalize `PATCH`, so doing this just to be consistent/safe.

www_src/lib/router.jsx
```diff
@@ -18,6 +18,13 @@ module.exports = {
           r = JSON.parse(hit);
         } catch (e) {}
       }
+    } else {
+      r = {
```
k88hudson: For local testing

www_src/lib/api.js
```diff
@@ -1,22 +1,24 @@
...
 module.exports = function (options, callback) {
   // Set default options
   defaults(options, {
-    method: 'get',
+    method: 'GET',
     useCache: false,
     json: {},
-    timeout: 60000    // 60 seconds
+    headers: {},
+    timeout: 60000 // 60 seconds
```
gvn: Maybe a little on the long side?

www_src/blocks/text.jsx
```diff
@@ -1,20 +1,53 @@
 var React = require('react');
-var utils = require('../lib/propUtils');
 var assign = require('react/lib/Object.assign');
+var Spec = require('../lib/spec');
```
thisandagain: I *really* like this new way of defining defaults for each block type.

www_src/components/range/range.jsx
thisandagain: Should this be throttled?

www_src/lib/api.js
thisandagain: I think for our expectations on 3G / 4G networks it is, but I'm worried about setting this two low for Edge networks in places like Bangladesh / China.

www_src/lib/spec.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+  });
+
+  return result;
+}
+
+function Spec(type, spec) {
+  if (typeof type !== 'string' || typeof spec !== 'object') throw new Error('Spec takes two params: type (String) and spec (Object)');
+  var error = getValidationErrors(spec);
+  if (error) throw new Error(error);
```
k88hudson: Note that if a spec is defined in a block and it doesn't pass validations (e.g. uses a category other than `styles` or `attributes`, see tests for more detail), it will throw

www_src/components/range/range.jsx
k88hudson: No, since this is only tied to a UI update and we want that to happen as quick as possible. Immediate state changes get cached via a `cacheEdits` function, which is only sent to the api on a `pause` event

www_src/lib/api.js
k88hudson: yeah, good call

www_src/lib/spec.js
```diff
@@ -0,0 +1,147 @@
...
+  };
+};
+
+Spec.propsToPosition = function (props) {
+  return {
+    position:'absolute',
+    top: 0,
+    left: 0,
+    transform: [
+      'translate('+props.x+'px, '+props.y+'px)',
```
thisandagain: Nitpick: These are pretty hard for me to read ... could we space this out a little bit?

www_src/pages/element/element.jsx
```diff
@@ -9,25 +9,69 @@ var editors = {
...
 render(React.createClass({
   mixins: [router],
   uri: function () {
-    var params = this.state.params
-    // for testing
-    if (!params.element) params = fakeParams;
-    return `/users/foo/projects/${params.project}/pages/${params.page}/elements/${params.element}`;
+    var params = this.state.params;
+    var element = params.element;
+    if (hash) {
```
thisandagain: This might be a little cleaner / faster done with an object literal instead of a switch.

k88hudson: aight

www_src/pages/element/element.jsx
```diff
@@ -9,25 +9,69 @@ var editors = {
...
   componentWillMount: function() {
-    api({uri: this.uri()}, (err, data) => {
-      this.setState({element: data});
+    this.load();
+  },
+  componentDidUpdate: function (prevProps) {
+    // resume
+    if (this.props.isVisible && !prevProps.isVisible) {
+      this.load();
+      console.log('restored!');
```
thisandagain: Stray `console.log`

www_src/pages/element/element.jsx
```diff
@@ -9,25 +9,69 @@ var editors = {
...
+    this.edits = edits;
+  },
+  save: function () {
+    var edits = this.edits;
+    var json = blocks[edits.type].spec.expand(edits);
+    api({method: 'patch', uri: this.uri(), json: {
+      styles: json.styles,
+      attributes: json.attributes
+    }}, (err, data) => {
+      console.log(err, data);
```
thisandagain: Stray `console.log`. Though we probably should log `if (err)`

www_src/pages/element/element.jsx
```diff
@@ -9,25 +9,69 @@ var editors = {
...
+      this.setState({
+        elements: edits
+      });
+      this.edits = false;
     });
   },
-  save: function (json) {
-    api({method: 'put', uri: this.uri(), json}, (err, data) => {
-      if (data) console.log('saved!');
```
thisandagain: Stray `console.log`

www_src/blocks/text.jsx
k88hudson: hurray

www_src/lib/spec.js
k88hudson: yes

www_src/pages/element/link-editor.jsx
```diff
@@ -11,11 +10,10 @@ var LinkEditor = React.createClass({
     React.addons.LinkedStateMixin
   ],
   getInitialState: function () {
-    var props = this.props.element || {};
-    return defaults(props, LinkBlock.defaults);
+    return LinkBlock.spec.flatten(this.props.element, {defaults: true});
   },
   componentDidUpdate: function () {
-    this.props.save(this.state);
+    this.props.cacheEdits(this.state);
```
thisandagain: :smile:

www_src/pages/make/make.jsx
```diff
@@ -1,32 +1,56 @@
...
+  },
+  load: function () {
     api({
-      uri: '/c0645e6953e9949f8e5c/raw/'
-    }, function (err, body) {
-      console.dir(err);
-      console.dir(body);
+      uri: '/users/1/projects'
+    }, (err, body) => {
+      if (err) return console.error('Error getting projects', err);
```
thisandagain: I wonder if we should do the `err` logging within the API module itself as to reduce this boilerplate and instead have a clean base for handling errors in the follow-up work. Perhaps even move the logs into the API module and replace all of these with `@todo`.

www_src/pages/page/page.jsx
```diff
@@ -31,13 +37,10 @@ var Page = React.createClass({
   },
 
   componentDidUpdate: function (prevProps) {
+    // resume
     if (this.props.isVisible && !prevProps.isVisible) {
       this.load();
       console.log('restored!');
```
thisandagain: Stray `console.log`

www_src/pages/make/make.jsx
k88hudson: Yeah that's a good idea. I still need to have the login to early return individually and whatnot though

www_src/pages/page/page.jsx
```diff
@@ -135,11 +145,20 @@ var Page = React.createClass({
...
-    this.setState({
-      elements: this.state.elements.concat([obj]),
-      showAddMenu: false
-    });
+  addElement: function(type) {
+    return () => {
+      var json = blocks[type].spec.generate();
+
+      api({method: 'post', uri: this.uri() + '/elements', json}, (err, data) => {
+        var state = {showAddMenu: false};
```
thisandagain: Can we wait to declare this `var` until we hit the success block?

www_src/pages/page/page.jsx
```diff
@@ -157,58 +176,66 @@ var Page = React.createClass({
...
 
-  addLink: function() {
-    this.appendElement(Generator.generateDefinition(Generator.LINK, {
-      href: "https://webmaker.org",
-      label: "webmaker.org",
-      active: false
-    }));
-  },
+    // Don't delete test elements for real;
+    if (parseInt(id, 10) <= 3) return alert('this is a test element, not deleting.');
```
thisandagain: What is the purpose of not allowing the deletion of test elements?

www_src/pages/page/page.jsx
k88hudson: as in, you think keeping the menu open is better in case of an error?

www_src/pages/page/page.jsx
```diff
@@ -157,58 +176,66 @@ var Page = React.createClass({
...
+      if (!data || !data.page) return console.log('Could not find the page');
+
+      var page = data.page;
+      var styles = page.styles;
+      var elements = page.elements.map(element => {
+        return this.flatten(element);
+      }).filter(element => element);
+      this.setState({
+        loaded: true,
+        styles,
```
thisandagain: Fancy ES6 :jeans:

www_src/pages/page/page.less
```diff
@@ -171,51 +171,55 @@
...
-        left: 0;
-        z-index: 2147483647; /* signed 32bit maxint */
-      }
+    .touchOverlay {
+      position: fixed;
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      z-index: 2147483647; /* signed 32bit maxint */
```
thisandagain: LOL

www_src/pages/page/page.jsx
k88hudson: Well... to be honest, it will stop the web view (with fake params) from being able to work. I was also going to use them to write some functional tests originally but never got around to it.

www_src/pages/page/page.jsx
thisandagain: Nope. You are right ... until we have an error handler, this looks good. :+1:

thisandagain:
:smile: :star2: :goat: :sheep: :smile: 

Amazingly *huge* effort. Some smaller comments / questions left in-line. Only one major issue that I see blocking this from landing and one smaller recommendation:

- A few projects (e.g. the second item in the discover list) seem to fail while loading which just presents the user with an infinite "Loading..." state. Could we simply remove these projects until error handling and loading states are integrated? Or ... is there something causing these to fail that needs to be resolved?
- We might want to specify a default / loading state for `Card` component as the list in the "Make" view is a little difficult to look at right now.

gvn:
This is working very well for me on device and browser! Nice!

thisandagain:
One other issue I've noticed is that there seems to be a perf issue when adding pages to a project.

k88hudson:
@thisandagain might related to the deleted_at bug thing, let's push up to api

k88hudson:
@thisandagain ok, i updated the API and the database, also added some blocking loading states on that view to resolve the issue

k88hudson:
@thisandagain also removed the empty/invalid proejcts

thisandagain:
Looks great. Only remaining blocker is the camera / gallery integration seems to be broken. It looks like the image is saved for a split second and then the webview seemingly reloads with Tucker.

k88hudson:
@thisandagain ok, let me take a look. the rebase was a bit hairy so I probably screwed up something there

k88hudson:
@thisandagain ok, this makes sense, the api update is getting triggered because bring up the camera is pausing the activity. will have to think about how to get around this

thisandagain:
Confirmed camera problem being caused by a duplicate webview being loaded "on-top" of the existing webview. This is good to merge once #1906 lands.

PR #1872, opened 2015-05-14
Crosswalk Impl & Notes

ryanw-se:
Crosswalk implementation for Android has been completed, its a full conversion from the old WebView. Because we have switched over to Crosswalk, some previously written JavaScript functionality will be broken by this update. These will need to be addressed before this can be accepted, as it essentially breaks the entire app.

@thisandagain and @k88hudson are looking into the JavaScript functionality that broke during the conversion, once that's completed and fixed this should be good to merge!

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ npm install
 #### Android
 While the majority of `webmaker-android` is built using Web technologies, it runs within a native Android wrapper that is included as part of this codebase. If you would like to make changes to the wrapper or if you'd like to test the app, we recommend you use [Android Studio](http://developer.android.com/sdk/index.html).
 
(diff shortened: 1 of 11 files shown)


ryanw-se:
@thisandagain We will probably need to have a discussion sometime tomorrow in IRC about this, so far with a few patches I've converted WebmakerWebView over to crosswalk and a lot of things have broken on the Javascript end.

thisandagain:
@ryanw-se Sounds good! Talk tomorrow. Want to send along a "WIP" (work-in-progress) pull request so I can see what you are running into?

thisandagain:
@ryanw-se Sounds good!

thisandagain:
Merged! :smile: :cake: :cat: :boom:

PR #1877, opened 2015-05-15
Button label editing

Pomax:
swaps a label for an input element for native text input while editing.

### DIFF
--- a/www_src/blocks/link.jsx
+++ b/www_src/blocks/link.jsx
@@ -10,10 +10,14 @@ var Link = React.createClass({
       borderRadius: 5,
       backgroundColor: '#69A0FC',
       fontFamily: 'Roboto',
+      whiteSpace: 'pre',
       innerHTML: 'Tap me',
       active: false
     }
   },
+  mixins: [
+    require('./textedit')
+  ],
   getDefaultProps: function () {
     return this.defaults;
   },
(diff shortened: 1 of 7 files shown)


flukeout:
@Pomax 

**Done button bug**
* Edit the button label
* Don't finish editing, but hit the back arrow until you are in the map view
* Edit the button again
* "Edit Text" button will still say "Done" - it should say "Edit Text"

Note: we have an issue to get rid of the "Done" button altogether, so might not be worth fixing. Maybe we just kill it off?


**Other Feedback**
* "Edit Text" should read "Edit Label"
* When I edit the label, the caret should be placed at the end of the text, not at the beginning
* Unlike the Text Element, if I tap the Button in the preview, I don't get to edit the text, can we add tap-to-edit behaviour to keep it consistent?

Pomax:
let's go will killing the button back off, then.
1. kk
2. are you sure? as an HTML element, this is the default behaviour. 
3. yeah that's odd, it should not handle the tap as a button press, but instead handle it as a text activation.

Pomax:
All the things should be addressed now, except for the button's existence

flukeout:
@Pomax 

A bug:
* Add a button
* Edit the label
* When editing place the caret in the middle of a word
* type a character
* caret jumps to the end of the label

Pomax:
all thee bugs should be gone

flukeout:
**Space bug**
* If you add spaces at the end of your label
* If you use the "Done" keyboard key or tap away from the field, they are trimmed
* If you tap the "Back" arrow, the spaces remain in the label

They should go away.

Pomax:
this seems to be an issue related to Android swapping out the view without the neccessary events triggering first (like the `onblur` event). @thisandagain might know how we can hook into that for some more flexibility.

The reason we can't `trim()` on every keystroke is that if we did, you'd never be able to type a sentence, so this is a really interesting bug that might not have a good solution for it without making Android hook into the view to say "yo, you're getting swapped out, do what you need to do right now and then I'm caching your state and getting rid of you"

flukeout:
Server side text trimming (as a service)

Pomax:
passing it over to @k88hudson for a code review before merging

thisandagain:
This is excellent. Thanks @Pomax and @flukeout ! R+

PR #1881, opened 2015-05-15
fix element mispositioning when going from project page preview to page

Pomax:
This fixes the whole nonsense with elements being positioned weirdly when going from pagae preview in the "project" view to the actual page editing view.

Consequence: elements are no longer generated in the middle of the page.


k88hudson:
This is indeed the right fix, although i am already doing this in https://github.com/mozilla/webmaker-android/pull/1869/files#diff-107e841757fc598ba7becb5966008e02L14 -- if you don't mind, i'd like to just stick with that

PR #1886, opened 2015-05-16
Selectnext

Pomax:
fixes https://github.com/mozilla/webmaker-android/issues/1885


flukeout:
@Pomax This seems good, but let's file an issue on what to do when we auto-select an element that is totally obscured by another element (lower z-index). As it stands, the user won't be able to see which element the trash can will affect.

Pomax:
filed followup as https://github.com/mozilla/webmaker-android/issues/1887

PR #1898, opened 2015-05-18
Don't allow dragging items outside of screen

### LINKED ISSUE #1789: Possible to drag elements off of the Page Editor in such a way that they are no longer visible and draggable
If you take an image, rotate it 45 degrees, and drag it up to the top right corner, you can drag it up and away far enough that it becomes invisible and if you leave it there, you won't be able to access it again in any way.

errietta:
Closes #1789

### DIFF
--- a/www_src/components/el/el.jsx
+++ b/www_src/components/el/el.jsx
@@ -94,14 +94,54 @@ var El = React.createClass({
   },
 
   handleTranslation: function(x, y) {
+    var thresh = 45;
+
+    var width = document.body.getBoundingClientRect().width/2 - thresh;
+    var height = document.body.getBoundingClientRect().height/2 - thresh;
+
     this.setState({
       x: x,
       y: y
     }, function() {
-      if(this.props.onUpdate) {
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks @errietta ! This is a great start ... when running locally I'm seeing an issue with larger elements (like photos from my camera) where I'm able to drag them up and down but not left and right. Also, elements should be able to be cropped by the bounding box of the page just not placed off the page entirely. We may want to specify a minimum number of pixels that have to be visible (40px?) before the element can no longer be moved.

thisandagain:
This is getting close!

errietta:
@thisandagain took out the x > width tests... still can throw stuff off, also now it handles horribly. HALP

errietta:
@thisandagain think it's better now, Re-R?

thisandagain:
This looks great! @k88hudson made some upstream changes that affected this however. Want to re-apply your fix?

thisandagain:
Awesome work @errietta . R+

Issue #1640, opened 2015-05-04
Build - Play View


vazquez:
Update Play View as per Luke's UX in #1566

vazquez:
Refining the Play View UI.

##### Play Map View

![play map view](https://cloud.githubusercontent.com/assets/1119821/7461583/15b13462-f260-11e4-9bb2-61bec296bfc3.png)  

##### Play Map View More Actions Menu

![play map view more actions](https://cloud.githubusercontent.com/assets/1119821/7461591/17f69258-f260-11e4-8fa8-860c1560fd15.png)  

##### Play Page View
After tapping the Zoom In button, the user enters the page specific play view. Notice that the header disappears (it slides up, out of view). We do, however, introduce a floating more actions button. 

![play page view](https://cloud.githubusercontent.com/assets/1119821/7461593/19b8b968-f260-11e4-83b9-4b6a15b11929.png)

##### Play Page View More Actions
When the user taps on the floating more actions button, they disclose the following menu.

![play page view more actions](https://cloud.githubusercontent.com/assets/1119821/7462278/1aead12c-f265-11e4-9563-5bfb1203c9b1.png)

cc: @xmatthewx @flukeout - Anything I'm missing?

thisandagain:
Why is the play UI orange? Seems disconnected from the visual language of the rest of the product.

vazquez:
Luke has been exploring in his prototypes having custom project colours (dark blue, orange, all black, white), but I can always bring it back to default blue.

thisandagain:
I don't think that is in scope for v1 ... let's keep it consistent for now.

flukeout:
@vazquez @thisandagain @xmatthewx Let's leave default blue for now. The play prototype simulates authors being able to set the header / UI color - but we'll tackle that later.

vazquez:
Changing to blue

vazquez:
Refining the Play View UI.

##### Play Map View

![play map view](https://cloud.githubusercontent.com/assets/1119821/7462681/54840ab8-f268-11e4-8151-ee8be5010d8d.png)


##### Play Map View More Actions Menu

![play map view more actions](https://cloud.githubusercontent.com/assets/1119821/7462683/57492062-f268-11e4-873a-44f7bf5b2d3c.png)


##### Play Page View
After tapping the Zoom In button, the user enters the page specific play view. Notice that the header disappears (it slides up, out of view). We do, however, introduce a floating more actions button. 

[...]

PR #1901, opened 2015-05-18
play mode

gvn:


www_src/pages/project/project.jsx
```diff
@@ -56,6 +56,13 @@ var Project = React.createClass({
     };
   },
 
+  // TODO - set isPlayOnly based on route
+  getDefaultProps : function() {
+    return {
+      isPlayOnly: false
```
gvn: Change to `true` to see Play Mode.

gvn:
Related to #1640

gvn:
Note that I didn't add a zoom out button since you can already pinch zoom. I tried it, but it felt a bit weird having 2 zoom methods.

thisandagain:
@flukeout Mind taking a look at this?

flukeout:
**Some feedback**
* Add a "Zoom Out" button when view is zoomed in to single page
* Add ability to swipe between adjacent pages when zoomed in
* Block "free panning" when zoomed in to one page
* When pinching out from the zoomed-in-to-page view the view jumps instead of animating to a lower zoom level
* Remove "selected" border from this view
* When zooming in to single page by tapping, change the header treatment to the one pictured earliner in this ticket
* A thought - the most zoomed in you can get with pinching seems a bit too close to the size of the page in the single-page view, what do you think?

@gvn What are your thoughts on these? Should we break them out into separate tickets? Ping me to chat when you've read these over.

gvn:
Kate, this addresses the stuff checked off above. R?

www_src/pages/project/project.jsx
```diff
@@ -53,7 +54,8 @@ var Project = React.createClass({
         x: 0,
         y: 0
       },
-      zoom: DEFAULT_ZOOM
+      zoom: DEFAULT_ZOOM,
+      isPageZoomed: false
```
k88hudson: Could you not maybe use `selectedEl` for this instead of a new property?

www_src/pages/project/project.jsx
```diff
@@ -139,22 +139,55 @@ var Project = React.createClass({
...
         if (!didMove) return;
 
-        var state = {camera: {
-          x: currentX,
-          y: currentY
-        }};
-        if (typeof currentZoom !== 'undefined') state.zoom = currentZoom;
-        this.setState(state);
-        startX, startY, startDistance, currentX, currentY, currentZoom = undefined;
+        if (!this.state.isPageZoomed) {
```
k88hudson: Can you maybe flip this around so it's a bit more readable and add a comment for the not selected state?

www_src/pages/project/project.jsx
```diff
@@ -331,21 +392,18 @@ var Project = React.createClass({
               page,
               selected: page.id === this.state.selectedEl,
               transform: this.cartesian.getTransform(page.coords),
-              onClick: this.selectPage(page)
+              onClick: isPlayOnly ? this.zoomToPage.bind(this, page.coords) : this.selectPage.bind(this, page)
```
k88hudson: i would put this logic in the this.selectPage handler -- you could rename it to `onPageClick` or something it that makes more sense

gvn: I feel kind of gross doing this, but otherwise the methods get immediately invoked on every render... :sob:

k88hudson:
Just a few suggestions, this is looking great

k88hudson:
@gvn also, i'm unable to select pages anymore in edit mode

www_src/pages/project/project.jsx
gvn: It's not really the same thing though. Zooming in on a page isn't "selecting" it.

www_src/pages/project/project.jsx
gvn: Added a comment, but there's no else case. :wink:

gvn:
Ok, should be all good now! @k88hudson

k88hudson:
R+ [[unflagged]](http://plzkthx.herokuapp.com)

PR #1905, opened 2015-05-19
Improve animation / transition on loading of webviews

thisandagain:
Really tiny detail, but helps with the perceived performance of loading new views onto the stack and fixes a rendering glitch on older versions of Android.


k88hudson:
R+

PR #1906, opened 2015-05-19
Fix critical issue where multiple webviews get added to the activity

thisandagain:
Fix critical issue where multiple webviews get added to the activity when a user hits 'back'.


k88hudson:
gooood catch

PR #1907, opened 2015-05-19
[Fix #1790] Tapping outside the bounds of an element deselects elements

### LINKED ISSUE #1790: Strange "tap page to deselect element" behaviour for Text Elements
You can deselect the currently selected element by tapping only roughly middle third (vertically) portion of an empty page in the Page Editor. 

To reproduce:
* On a blank page, add a text element
* Tap it to select it
* Tap above it on the empty page near the top, nothing happens
* Tap below it on the empty page near the top, nothing happens
* Tap next to it on the empty page (or anywhere in roughly the middle 1/3 of the page) -> element becomes deselected

Let's make it so that tapping anywhere that isn't the selected element deselects it.

k88hudson:


thisandagain:
Nice! R+

PR #1908, opened 2015-05-19
[Fix #1851] Hide the text alignment editor until multiline text is added back

### LINKED ISSUE #1851: Remove "Text Align" option from Text Element
Currently, we don't support multi-line text, which means that changing the text alignment of the Text Element has no noticeable effect on the Text Element when it's on a Page, even though it changes what you see in the Element Editor.

k88hudson:


thisandagain:
R+

PR #1914, opened 2015-05-19
[Fix #1912] Project view crashes if state.params is undefined

### LINKED ISSUE #1912: Drag to pan is broken on first load
**Repro Steps:**
1. Go to discover tab
2. Tap 1st project
3. Try to pan

**Expected:** Page map pans as you drag it.
**Actual:** Nothing happens when you drag.

k88hudson:


gvn:
I think that fixed it!

PR #1915, opened 2015-05-19
Hook up 'play' activity. Re GH-1640

thisandagain:


gvn:
Looks like you still need to add the prop to Project to indicate play mode.

See: https://github.com/mozilla/webmaker-android/pull/1901/files#diff-02fce5293d9817377f211ad1dde7633dR62

thisandagain:
This gives you `mode: play` on the router mixin as per https://github.com/mozilla/webmaker-android/blob/develop/README.md#router

gvn:
Ok cool. Is there a web route for this or can you only access this view on device?

thisandagain:
It's just linked to the "Project" view from your PR.

gvn:
Ok, I'm seeing it via remote inspector. I wonder if we should add `mode: "edit"` for edit mode.

thisandagain:
Ok. All set.

PR #1916, opened 2015-05-19
Build configuration fixes

thisandagain:
Should resolve issues for x86-based emulators

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,17 +24,12 @@ android {
         targetCompatibility JavaVersion.VERSION_1_7
     }
 
-    productFlavors {
-        armv7 {
-            ndk {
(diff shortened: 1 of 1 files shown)


k88hudson:
I don't see that error anymore. I had to build a couple of times before it worked (I got `Session 'app': error` with no other output) but might have been an unrelated issue.

Tested on an x86 emulator/API 21/5.0.1

thisandagain:
Generally when the build config changes it's a good idea to clean the project. That might have been what caused that particular error.

k88hudson:
@thisandagain I did, but who knows ^_^

thisandagain:
Gah. Let me know if you see it again. Thanks @k88hudson

PR #1923, opened 2015-05-20
Fix #1882 - Use centre as the origin

### LINKED ISSUE #1882: Button position shifts unexpectedly after changing the label.
Say I'm going to edit this button:
image

I'm going to change the label to something longer. The element stays centered in the preview...

image

image

But when I return to the Page, the suggested behaviour in the Element preview makes this positioning a little unexpected:

image

**So, two bad things**
* I feel like the button should have stayed centered (so we track the x,y position from the middle of the button not the top left)
* The Button Label shouldn't have wrapped, it should stay one line and run off of the page

k88hudson:
* since this requires some complex positioning wrapper divs, I refactored both the page and project views to use an `Element-Group` and `El` component
* `El` is named as such to avoid confusion when debugging with the `Element` page component
* Warning! This will totally mess up existing coordinates. :warning:


k88hudson:
R in person by @gvn :sunglasses:

PR #1931, opened 2015-05-20
[Fix #1834] Override back button

### LINKED ISSUE #1834: Surprising behaviour when tapping back arrow on Add Element screen
When you are in the add Element View, tapping the back arrow takes you back to the map view. Tapping anywhere else on the page just dismisses it and returns you to the Page Editor.

This behaviour seems slightly surprising since I don't think of myself as being in the Page Editor but rather in the Add Element view. Should the back arrow just cancel the Add Element view?

@vazquez Thoughts?

k88hudson:


app/src/main/java/mozilla/org/webmaker/activity/Page.java
```diff
@@ -7,4 +7,9 @@
     public Page() {
         super("page", R.id.page_layout, R.layout.page_layout, R.menu.menu_page);
     }
+
+    @Override
+    public void onBackPressed() {
```
k88hudson: I could also move this `WebmakerActivity` to be more generic, but I figured it might be better to not mess around with the back button where not necessary. What do you think?

thisandagain: We've got a bunch of places where we need to override the back button ... moving it into `WebmakerActivity` I think is a good idea.

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -82,6 +84,11 @@ public void logText(String txt){
         Log.v("wm", txt);
     }
 
+    @JavascriptInterface
+    public void goBack() {
+        mActivity.finish();
```
thisandagain: This should likely call `onBackPressed` to make it more generic / work everywhere.

app/src/main/java/mozilla/org/webmaker/view/WebmakerWebView.java
```diff
@@ -25,6 +25,6 @@ public WebmakerWebView(Context context, Activity activity, String pageName, JSON
         this.load("file:///android_asset/www/pages/" + pageName + "/index.html", null);
         this.setResourceClient(new WebClient(this));
         this.setBackgroundColor(getResources().getColor(R.color.light_gray));
-        this.addJavascriptInterface(new WebAppInterface(context, routeParams), "Android");
+        this.addJavascriptInterface(new WebAppInterface(context, activity, routeParams), "Android");
```
thisandagain: You don't need to pass along the `Activity` here. You can cast the context instead:
```java
Activity activity = (Activity) context;
```

thisandagain:
This looks good! Just a few comments.

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
k88hudson: Unfortunately I can't do that, since `onBackPressed` is being overridden, right?

app/src/main/java/mozilla/org/webmaker/view/WebmakerWebView.java
k88hudson: ah cool!

thisandagain:
Awesome. R+

PR #1938, opened 2015-05-20
[Fix #1930] Add loading state to create project

### LINKED ISSUE #1930: Exiting from a the Project Editor returns you to a "loading state" instead of the Make view.
Steps to reproduce
* In Make view, create a new project
* Once the project loads, his the back arrow
* You will return to a view with a progress bar under the Project header

You should return immediately to the the Make view.

### LINKED ISSUE #1932: [MIGRATED] Create Project button is too picky about what kinds of taps it likes
It's very easy to tap the Create Project button on the Make view with no effect. It seems like it only likes very quick taps, holding down your finger for even a short period of time before releasing makes the button unresponsive.

Let's not set a limit (or at least not as aggressive) to how fast the tap needs to be, especially because the "pressed state" on the button suggests that you're using it correctly even if holding it for a long time.

For reference, the Make and Discover tabs can be tapped and held for a long time before releasing your finger, and they still work.

k88hudson:
Should also resolve #1932

### DIFF
--- a/www_src/pages/make/make.jsx
+++ b/www_src/pages/make/make.jsx
@@ -3,12 +3,14 @@ var render = require('../../lib/render.jsx');
 var api = require('../../lib/api.js');
 var Card = require('../../components/card/card.jsx');
 var Link = require('../../components/link/link.jsx');
+var Loading = require('../../components/loading/loading.jsx');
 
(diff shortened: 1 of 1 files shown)


gvn:
R+

PR #1956, opened 2015-05-20
[Fix #1948] Update positioning based on id not index

### LINKED ISSUE #1948: Changing element properties often causes all other elements to reposition
* Add 4 toucans
* Add 4 buttons
* Space them out all over the page
* Change any attribute of one of the toucans
* A lot of the elements on the page will shift in position

k88hudson:
Because using index is pretty unsafe


gvn:
R+
 [[unflagged]](http://plzkthx.herokuapp.com)

dog-headbang

PR #1978, opened 2015-05-21
[Fix #1973] Fix loading issue on Make view

### LINKED ISSUE #1973: When I return to Make view from creating a new project, the page view is stuck loading
**Steps to reproduce**
* Create a new project
* Immediately go back to the Make view
* The Make view has a loading bar below the title
  * It doesn't go away
  * I can't create a new project or open an existing one while it's loading

Note: I can pull-to-refresh to get out of this state (but we're removing that from this view anyway)

k88hudson:
This is semi-critical

### DIFF
--- a/www_src/pages/make/make.jsx
+++ b/www_src/pages/make/make.jsx
@@ -63,25 +63,22 @@ var Make = React.createClass({
       if (err) {
         return this.onError(err);
       }
-      
       if (!body || !body.project) {
         return this.onEmpty();
       }
-      
       if (window.Android) {
         window.Android.setView('/projects/' + body.project.id);
-      } else {
-        this.setState({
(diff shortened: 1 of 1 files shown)


www_src/pages/make/make.jsx
```diff
@@ -63,25 +63,22 @@ var Make = React.createClass({
...
-            id: body.project.id,
-            title: defaultTitle,
-            thumbnail: {},
-            // todo
-            author: userInfo
-          }].concat(this.state.projects)
-        });
       }
+      this.setState({
+        loading: false,
```
k88hudson: this should happen in all cases, not just if Android doesn't exist

thisandagain:
R+ :+1:

PR #1979, opened 2015-05-21
Resolve lint errors and remove universal (FAT) APK from build

thisandagain:
Be sure to test this against both an x86-based emulator as well as an ARM-based device.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,13 +1,13 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 21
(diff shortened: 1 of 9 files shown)


thisandagain:
@gvn @ncalexan R? [[flagged!]](http://plzkthx.herokuapp.com)

thisandagain:
@ncalexan Any thoughts on how to further improve build times would be most appreciated. :smile:

gvn:
R+, this seems to shave off about 20 seconds!

Issue #1899, opened 2015-05-18
Page - Default background color


LauraReynal:
@thisandagain @xmatthewx

thisandagain:
:yellow_heart: :yellow_heart: :yellow_heart: :yellow_heart: :stuck_out_tongue_winking_eye:

thisandagain:
Because we won't have the background color editor in v1.0.0 we should select a more neutral default.

LauraReynal:
:yellow_heart: :sun_with_face: :sunflower: :sunny: :sunrise: :sunglasses:

vazquez:
What about a very subtle blueish white? `#FAFCFF`


![tile view copy 2](https://cloud.githubusercontent.com/assets/1119821/7752628/7a22d01c-ffae-11e4-8a74-58ac6c063aca.jpg)

thisandagain:
Oooh ... that's nice. :+1:

PR #1980, opened 2015-05-21
Replace yellow page background with off-white. Resolves GH-1899

thisandagain:
:white_circle: :white_large_square: :white_circle: :white_large_square: :white_circle: :white_large_square:


gvn:
ibs

PR #1991, opened 2015-05-22
refactoring text input to allow overflow

### LINKED ISSUE #1879: maxlength not working on TextInput
This seems to be a bug in Android's webview implementation, but `maxlength` isn't being enforced on device.

**Repro Steps:**
1. Type more than the character limit

**Expected:** Once limit is reached, any further character input is blocked.
**Actual:** Additional characters can be entered beyond the limit.

gvn:
Closes #1879


thisandagain:
Nice and simple. R+

Issue #1982, opened 2015-05-21
Page / Element - Strange behavior when moving between views


flukeout:
Steps to reproduce
* Create a new project
* Edit the start page
* Add a toucan
* Make it huge!!
* Go back to map view
* Page thumbnail is correct
* Edit the page again
* Make the toucan small
* Go back to map view
* Toucan still appears large in page thumbnail

It should reflect the new, smaller size. If you add a new element to the page, then the thumbnail updates, but often to an inaccurate representation of the page.

thisandagain:
I'm seeing this fairly consistently (8/10 tries this morning running latest on Nexus 5 / 5.1.1) throughout all elements, not just images. Marking as critical.

k88hudson:
ok, on it

thisandagain:
I'm still able to reproduce the updating issue, but also seems to cause or amplify another issue in which if I add multiple items it starts to get into a strange loop in which new elements won't be displayed or update. Perhaps a race condition or even an issue where we are using brittle IDs or assigning properties to the wrong element? At the very end of the screen recording you'll notice that a text element just disappears (I did not delete it). Screen recording: https://www.dropbox.com/s/vapc2kysvoid36w/bug.mp4?dl=0

Pomax:
blocked by https://github.com/mozilla/webmaker-android/issues/2009

PR #1994, opened 2015-05-22
making thumbnails update

gvn:
Not sure if this is terrible, but it seems to work.

### DIFF
--- a/www_src/components/el/el.jsx
+++ b/www_src/components/el/el.jsx
@@ -70,7 +70,7 @@ var El = React.createClass({
   },
 
   render: function() {
-    var style = Spec.propsToPosition(this.state);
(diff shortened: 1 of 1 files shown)


gvn:
Closes #1982

thisandagain:
I'm still able to reproduce the updating issue, but also seems to cause or amplify another issue in which if I add multiple items it starts to get into a strange loop in which new elements won't be displayed or update. Perhaps a race condition or even an issue where we are using brittle IDs or assigning properties to the wrong element? At the very end of the screen recording you'll notice that a text element just disappears (I did not delete it). Screen recording: https://www.dropbox.com/s/vapc2kysvoid36w/bug.mp4?dl=0

thisandagain:
Landing this as an incremental improvement. Updating #1982 with details.

Issue #2007, opened 2015-05-26
Refactor WebAppInterface for performance / clarity


thisandagain:
The current `WebAppInterface` class exposes JS methods for a variety of things that share the same purpose (persisting state). This should be refactored to use the new `LRUCache`-based `MemStorage` class where reasonable.

ryanw-se:
@thisandagain I'll take care of this tomorrow, if it hasn't already been assigned to someone!

thisandagain:
Thanks @ryanw-se but I'm almost done with this one! :smile:

Issue #1670, opened 2015-05-06
Set Link destination (within a project)


xmatthewx:
Implement the feature to allow users to pick a page within their project as the destination for their link. 

 1. In link element editor, user clicks set destination
 1. Map view appears with icon on current page
 1. User pans and taps destination page
 1. Icon appears on destination with line connecting to origin
 1. User taps checkmark in top right
 1. Return to element editor

Note
 - Unset link: When an unset link is selected in page editor, it  includes a button to set destination. 
 - Set link: When a link is selected in page editor, it includes a button to follow the link
 - Play link: In Play view, the button behaves like button. Yay.

UX is complete. UI in progress in #1549

xmatthewx:
cc: @flukeout @vazquez for review

### Element on page

 - Button is disabled to allow select and transforms
 - Pencil opens element editor
 - `Set Link Destination` goes to page selector

![button add to page](https://cloud.githubusercontent.com/assets/1119821/7504843/26721af2-f402-11e4-89e2-be43a474027b.png)


## Link editor

 - Users can set destination and button style.
 - Tap button in preview to edit the string. 
 - Tap checkmark in header to confirm changes.
 - Nice to have: scroll overflow behind the preview.

![button options](https://cloud.githubusercontent.com/assets/1119821/7505199/5a1a1c8a-f405-11e4-869a-1ed5d026d1a4.png)

## Pick Destination Page

Users arrive here after tapping `Set Destination`

 1. Show page selector
   - center map view on the page that contains the button
   - mark this page with placemarker icon
 2. User taps destination
  - add selected state to page
  - add destination flag icon
  - confirmation checkmark appears
  - nice to have: draw connecting line between pages
 1. User confirms changes
 1. Go to element editor

![button pick destination page](https://cloud.githubusercontent.com/assets/1119821/7505333/4f7b101c-f406-11e4-8654-1e0d35715ebc.png)  

![button set destination page](https://cloud.githubusercontent.com/assets/1119821/7505335/52f8eb24-f406-11e4-9634-e995881ffa6a.png)

## Destination Set

Button helper now reads `Follow Link`

![button add to page link exists](https://cloud.githubusercontent.com/assets/1119821/7505564/cabe02d2-f408-11e4-9ebb-23ff9146f1a4.png)

flukeout:
Looks good! What about ```Change Link Destination``` in the element editor after a page has been set.

xmatthewx:
[...]

PR #2010, opened 2015-05-27
Refactor Android <-> JS storage layer

thisandagain:
Resolves GH-2007
Unblocks GH-1670

### DIFF
--- a/app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
+++ b/app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
@@ -33,110 +33,47 @@ public class WebAppInterface {
         mActivity = (BaseActivity) context;
         mPrefKey = "::".concat(mContext.getClass().getSimpleName());
         mPrefs = mContext.getSharedPreferences(mPrefKey, 0);
-        mPageState = mPrefs.getString("page_state", "{}");
         mRoute = routeParams;
         Log.v("wm", "getting state " + mPrefKey + ": " + mPageState);
     }
 
     /**
      * ---------------------------------------
-     * SharedPreferences
+     * Disk-based Storage
      * ---------------------------------------
      */
 
     @JavascriptInterface
-    public String getSharedPreferences(String key, boolean scope) {
+    public String getSharedPreferences(String key, final boolean scope) {
         SharedPreferences getter = mContext.getSharedPreferences(WEBMAKER_PREFS, 0);
-        if (scope) {
-            key = key.concat(mPrefKey);
-        }
+        if (scope) key = key.concat(mPrefKey);
         return getter.getString(key, null);
     }
 
     @JavascriptInterface
-    public void setSharedPreferences(String key, String value, boolean scope) {
+    public void setSharedPreferences(String key, final String value, final boolean scope) {
         SharedPreferences.Editor editor = mContext.getSharedPreferences(WEBMAKER_PREFS, 0).edit();
-        if (scope) {
-            key = key.concat(mPrefKey);
-        }
+        if (scope) key = key.concat(mPrefKey);
         editor.putString(key, value);
         editor.apply();
     }
 
-    @JavascriptInterface
-    public String getState() {
-        return mPrefs.getString("page_state", "{}");
-    }
-
-    @JavascriptInterface
-    public void setState(String serializedState) {
-        SharedPreferences.Editor edit = mPrefs.edit();
-        edit.putString("page_state", serializedState);
-        edit.apply();
-    }
-
     /**
      * ---------------------------------------
-     * Utility
+     * Memory-based Storage
      * ---------------------------------------
      */
 
     @JavascriptInterface
-    public void logText(String txt){
-        Log.v("wm", txt);
+    public String getMemStorage (String key, final boolean scope) {
+        if (scope) key = key.concat(mPrefKey);
+        return MemStorage.sharedStorage().get(key);
     }
 
     @JavascriptInterface
-    public void goBack() {
-        mActivity.runOnUiThread(new Runnable() {
-            public void run() {
-                mActivity.goBack();
-            }
-        });
-    }
-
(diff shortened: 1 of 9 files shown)


thisandagain:
@gvn and @Pomax R?

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -118,7 +129,7 @@ public void run() {
     public void setView(final String url, final String routeData) {
         Activity activity = (Activity) mContext;
         if (activity == null) return;
-        MemStorage.sharedCache().put(url, routeData);
+        MemStorage.sharedStorage().put("route::data", routeData);
```
Pomax: let's pack that string up as a constant or enum

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -128,35 +139,20 @@ public void run() {
...
     }
 
     @JavascriptInterface
-    public void getFromMedia() {
-        Element elementActivity = (Element) mContext;
-        if (elementActivity != null) {
-            elementActivity.dispatchMediaIntent();
-        }
+    public String getRouteData() {
+        return MemStorage.sharedStorage().get("route::data");
```
Pomax: magic string instance #2. Definitely needs packing up.

app/src/main/java/mozilla/org/webmaker/storage/MemStorage.java
```diff
@@ -1,61 +1,20 @@
...
 
-import android.app.ActivityManager;
-import android.content.Context;
 import android.util.LruCache;
 
 public class MemStorage {
 
-    private static Context context;
-    private static int cacheSize = 5 * 1024 * 2014;
+    private static int cacheSize = 24 * 1024 * 2014;
```
Pomax: what's this new number based on? (we might want to document that in the code so we know what it should reflect and whether it's gone stale)

www_src/lib/api.js
```diff
@@ -54,11 +54,11 @@ module.exports = function (options, callback) {
...
     // Set cache if window.Android is available
     if (window.Android) {
-      window.Android.setSharedPreferences(key, JSON.stringify(body), false);
+      window.Android.setMemStorage(key, JSON.stringify(body), false);
     }
 
     // If there is a callback, forward the response body
-    if(callback) {
+    if (callback) {
       callback(null, body);
```
Pomax: nit: there is no good reason to ever use `null` in JS context. Can we make this "false" if we're editing this file anyway?

www_src/lib/router.jsx
```diff
@@ -30,9 +30,22 @@ module.exports = {
 
     return r;
   },
+  getRouteData: function () {
+    var r = {};
```
Pomax: nit: minification will take care of `r`, let's call it `routes` or something in the full code?

Pomax:
question: since the vast majority of the time memCache is going to be scoped, does it make sense to have a "scope" boolean, rather than a "global" boolean? And then have memCache be automatically scoped unless explicitly unscoped?

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -33,59 +33,76 @@ public WebAppInterface(Context context, JSONObject routeParams) {
...
     /**
      * ---------------------------------------
-     * SharedPreferences
+     * Disk-based Storage
      * ---------------------------------------
      */
 
     @JavascriptInterface
-    public String getSharedPreferences(String key, boolean scope) {
+    public String getSharedPreferences(String key, final boolean scope) {
```
Pomax: good idea to add a functor overload with a default, like
```
public String getSharedPreferences(String key) {
  return getSharedPreferences(key, true);
}
```

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -33,59 +33,76 @@ public WebAppInterface(Context context, JSONObject routeParams) {
...
-        if (scope) {
-            key = key.concat(mPrefKey);
-        }
+        if (scope) key = key.concat(mPrefKey);
         return getter.getString(key, null);
     }
 
     @JavascriptInterface
-    public void setSharedPreferences(String key, String value, boolean scope) {
+    public void setSharedPreferences(String key, final String value, final boolean scope) {
```
Pomax: same overloading note

app/src/main/java/mozilla/org/webmaker/javascript/WebAppInterface.java
```diff
@@ -33,59 +33,76 @@ public WebAppInterface(Context context, JSONObject routeParams) {
...
+    /**
+     * ---------------------------------------
+     * Memory-based Storage
+     * ---------------------------------------
+     */
+
     @JavascriptInterface
-    public String getState() {
-        return mPrefs.getString("page_state", "{}");
+    public String getMemStorage (String key, final boolean scope) {
```
Pomax: ... etc =)

Pomax:
Runtime-wise I don't see anything blowing up, although the proliferation of makes (due to everyone making with the same user account) is definitely interfering with detailed testing.

thisandagain:
Thanks for the great review @Pomax . Feedback addressed in the latest commit.

Re-scope vs. global. It's fairly split right now and my guess is that we'll increase the sharing of objects between views (scope = false) over time as we optimize performance – thus the initial "scope" naming.

thisandagain:
That said, I think "global" is easier to understand. Will revise.

gvn:
Looks good. It would be nice to have a browser polyfill for dev outside of Android, but I'd say a follow up ticket for that is fine since it's a nice-to-have for developer ergonomics.

R+!

dance2

PR #2025, opened 2015-05-27
Change the element list for Page containers to a keyed object

### LINKED ISSUE #1982: Page / Element - Strange behavior when moving between views

Pomax:
Fixes #1982 by changing the list of elements from a blind array to a an
object that is keyed on element.ids instead.This lets us track the
currentElement by their element.id, too, rather than relying on any particular
array ordering (which may be invalidated by database storage and retrieval)

### DIFF
--- a/www_src/components/element-group/element-group.jsx
+++ b/www_src/components/element-group/element-group.jsx
@@ -5,42 +5,41 @@ var assign = require('react/lib/Object.assign');
 var ElementGroup = React.createClass({
   getDefaultProps: function () {
     return {
-      elements: [],
-      currentElement: -1,
+      elements: {},
+      currentElementId: -1,
       interactive: false
     };
   },
   render: function () {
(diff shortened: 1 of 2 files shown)


thisandagain:
Still missing one more semicolon.

thisandagain:
This is working really well locally. Confirmed fix for #1982

thisandagain:
R+ once lint issue is resolved:
```bash
www_src/components/element-group/element-group.jsx: line 20, col 44, Missing semicolon.
```

PR #2027, opened 2015-05-27
setting link destination in Project view w. hardcoded source page

gvn:


gvn:
@thisandagain This is more progress toward #1670. I'm trying to keep PRs in smaller chunks...also I just had a rather nasty rebase. :rage1: 

So far this only works in browser and relies on a hardcoded value for the "source page".

I don't think landing this will interfere with any on device functionality. We may or may not want to take out my new router fallback params though...

www_src/pages/project/project.jsx
```diff
@@ -278,6 +306,10 @@ var Project = React.createClass({
       state.camera = this.cartesian.getFocusTransform({x: 0, y: 0}, this.state.zoom);
     }
     this.setState(state);
+
+    if (this.state.params.mode === 'linkDestination') {
+      this.highlightPage('408', 'source'); // TEMP / TODO - Hardcoded. Should come from passed value.
```
gvn: Note this hardcoded page ID...works as of now, but if the project loses this page you might need to set another ID to see this working...

www_src/lib/router.jsx
```diff
@@ -20,6 +20,8 @@ module.exports = {
       }
     } else {
       params = {
+        // edit, play, linkDestination
+        mode: 'linkDestination',
```
gvn: `edit` might be a more sensible default...?

thisandagain: :+1:

www_src/pages/project/project.jsx
```diff
@@ -208,11 +213,33 @@ var Project = React.createClass({
...
-    this.setState({
-      camera: this.cartesian.getFocusTransform(el.coords, this.state.zoom),
-      selectedEl: el.id
-    });
+  /**
+   * Highlight a page in the UI and move camera to center it
+   * @param  {Number|String} id ID of page
+   * @param  {Number|String} type Type of highlight ("selected", "source")
+   */
+  highlightPage: function (id, type) {
```
thisandagain: Nice. I like moving this to a self-contained method.

www_src/pages/project/project.jsx
```diff
@@ -278,6 +305,10 @@ var Project = React.createClass({
       state.camera = this.cartesian.getFocusTransform({x: 0, y: 0}, this.state.zoom);
     }
     this.setState(state);
+
+    if (this.state.params.mode === 'linkDestination') {
```
thisandagain: This could use a line comment.

www_src/static/img/oval-droplet.svg
```diff
@@ -0,0 +1,22 @@
+<svg width="52" height="52" viewBox="0 0 52 52" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
thisandagain: Would be fine to put this through SVGO

www_src/static/img/oval-flag.svg
```diff
@@ -0,0 +1,22 @@
+<svg width="52" height="52" viewBox="0 0 52 52" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
thisandagain: Same here

thisandagain:
This looks good. No major blockers from my perspective. Just a few small notes and then R+

www_src/static/img/oval-droplet.svg
gvn: I actually did, just pretty printed. :poodle:

www_src/pages/project/project.jsx
gvn: Yeah, I tried it as 2 methods at first, which was more single responsibility-ish, but way less DRY...so...yeah. :cactus:

PR #2028, opened 2015-05-27
unsymlinked the hintrc

### LINKED ISSUE #2026: .jshintrc cannot be a symlink if we want to support devs on windows
symlinks are great on unix, but don't work on windows: windows just sees a text file with a relative file location in it, which breaks `npm test` on windows when it tries to load the jshintrc file and just sees "text" instead of JSON. We can, however, just point to the node_modules/mofo-style version directly.

Pomax:
fixes #2026

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,1 +1,0 @@
-node_modules/mofo-style/linters/.jshintrc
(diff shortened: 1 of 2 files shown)


thisandagain:
Confirmed in my Windows 8 VM.

Issue #1909, opened 2015-05-19
Persist project settings


gvn:
We'll need to hook up the Project Settings view to the new backend for persistence.

secretrobotron:
Just pinging here. @LauraReynal and I just ran into this.

secretrobotron:
Blocks #1665

Issue #1966, opened 2015-05-21
Implement loading states for projects, lists, and pages


vazquez:
We need an indeterminate loader for actions that have API calls such as adding an element, deleting a page, etc. This loader will show when network connections are slow. Otherwise for fast connections, this will rarely be shown.

cc / @k88hudson

vazquez:
#### Prototype

http://codepen.io/rvazquez/pen/xGELVZ?editors=110

Note: Use Firefox

What do you think @k88hudson, would this work?

k88hudson:
@vazquez looks great!

vazquez:
:rocket:

k88hudson:
Resolved by https://github.com/k88hudson/webmaker-android/commit/b82342ddaafd8bee5b9c2baa7eb99b5e38fdc371

PR #2034, opened 2015-05-28
Hook-up project settings

thisandagain:
Resolves GH-1909

### DIFF
--- a/www_src/components/text-input/text-input.jsx
+++ b/www_src/components/text-input/text-input.jsx
@@ -31,14 +31,14 @@ var TextInput = React.createClass({
   },
   render: function () {
     var linkState = this.props.linkState || this.linkState;
-    var valueLink = this.valueLink = linkState(this.props.id);
+    this.valueLink = linkState(this.props.id);
 
     return (
       <div className={'text-input' + (this.state.isTooLong ? ' maxed' : '')}>
         <label>{this.props.label}</label>
 
         <input
-          value={valueLink.value}
+          value={this.valueLink.value}
           ref="input"
           onChange={this.onChange}
           type="text"/>
--- a/www_src/components/text-input/text-input.jsx
+++ b/www_src/components/text-input/text-input.jsx
@@ -1,6 +1,7 @@
-var React = require('react');
+var React = require('react/addons');
 
 var TextInput = React.createClass({
+  mixins: [React.addons.LinkedStateMixin],
   propTypes: {
     maxlength: React.PropTypes.number.isRequired,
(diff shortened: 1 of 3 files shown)


thisandagain:
One note is that while this does get persisted to the DB, an API bug makes viewing the updated project titles in the make view difficult: https://github.com/mozilla/api.webmaker.org/issues/70

www_src/components/text-input/text-input.jsx
```diff
@@ -28,12 +30,15 @@ var TextInput = React.createClass({
     }
   },
   render: function () {
+    var linkState = this.props.linkState || this.linkState;
+    var valueLink = this.valueLink = linkState(this.props.id);
```
gvn: Does a local var need to be created for some reason? Seems like just using this.valueLink should be fine...

www_src/pages/project-settings/project-settings.jsx
```diff
@@ -1,21 +1,83 @@
...
+  getInitialState: function () {
+    return {
+      title: ''
+    };
+  },
+  componentWillMount: function () {
+    var _this = this;
+
+    // Build up URI for API requests
+    var params = this.state.params;
```
gvn: Since this is only used once, maybe just use `this.state.params` below...

www_src/components/text-input/text-input.jsx
thisandagain: :+1:

www_src/pages/project-settings/project-settings.jsx
```diff
@@ -1,21 +1,83 @@
...
       </div>
     );
+  },
+
+  /**
+   * Click handler for persisting changes to project settings.
+   *
+   * @return {void}
+   */
+  onDoneClick: function () {
```
gvn: You might consider adding logic to see if a PATCH is already being sent to avoid a bunch of stacked requests if someone taps a bunch of times (or even double taps for some reason...).

thisandagain: Great catch though I wonder if we should just deal with that globally as per #1966

gvn: Yeah might wanna do debouncing at the API level...

gvn:
kewl

R+

www_src/pages/project-settings/project-settings.jsx
thisandagain: Agreed. I think we should have a UI binding into `lib/api.js` rather than do this ad-hoc everywhere.

PR #2039, opened 2015-05-29
fixes trash icon location and existence if page count < 2

### LINKED ISSUE #2017: Immediately deleting first tile breaks project
1. Start new project
2. Highlight initial, existing tile
3. Delete it

Screen turns grey; can't add any tiles. Forced to go back to gallery.

### LINKED ISSUE #2029: move "project" trash icon to be in line with the "page" icon location
Project has the trash icon on the right of the (+), page has it to the left, which will probably end up having unintentional effects.

Pomax:
- fixes #2029 - trash icon for pages is now on the left, which aligns it with what we have on the dedicated page view
- fixes #2017 - only shows the trash icon if there is more than one page, so that you can't delete pages from projects with only a single page in them (which would leave you with an empty, useless, project view)

### DIFF
--- a/www_src/pages/project/project.jsx
+++ b/www_src/pages/project/project.jsx
@@ -467,6 +467,10 @@ var Project = React.createClass({
       }
     }
 
+    var removePageButton = this.state.pages.length > 1 ? (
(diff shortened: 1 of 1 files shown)


thisandagain:
R+ :+1:

PR #2041, opened 2015-05-29
proposal for global dispatcher w. link destination example case

gvn:


gvn:
Ok, this is a very rudimentary approach, but it works and feels easy enough to use.

I'm just using DOM events, but we could use whatever, and let dispatcher be a facade for any implementation. One nice thing we could do would be to attach analytics calls to the dispatcher, but I digress...

To see this in action, run http://localhost:4242/pages/page/ and click a button, then Set Destination.

gvn:
@thisandagain @Pomax @k88hudson

www_src/lib/dispatcher.js
```diff
@@ -0,0 +1,14 @@
...
+  fire: function (eventName, eventData) {
+    var customEvent = new CustomEvent(eventName, {
+      detail: eventData
+    });
+
+    window.dispatchEvent(customEvent);
+  },
+  on: function (eventName, callback) {
+    window.addEventListener(eventName, function(e) {
+      callback.call(this, e.detail);
```
Pomax: does it matter that the context is now the dispatcher instead of the window? They're both going to be something not related to the thing the event was fired by?

gvn: We could have it be `window` instead. I suppose that would be marginally better. :)

PR #2047, opened 2015-05-29
More progress on Linking

gvn:
Related to #1670


www_src/lib/router.jsx
```diff
@@ -20,7 +20,7 @@ module.exports = {
       }
     } else {
       params = {
-        mode: 'play', // 'edit', 'play', 'linkDestination'
+        mode: 'link', // 'edit', 'play', 'link'
```
thisandagain: Mind if we keep this default on `play`?

gvn: Whoops...yes, let me fix that.

thisandagain:
Great progress ... feels like this is starting to come together. :smile: 
R+

PR #2048, opened 2015-05-30
prevent weird running text

Pomax:
This restricts the `<input>` width to 100%, and sets the "in editor" span width to 100% with an ellipsis to hint at more content. When editing, the user can swipe to move to the content like in any other app, and in the page view, there is no cutoff so all the content is visible.


thisandagain:
R+ :panda_face:

PR #2049, opened 2015-06-01
Don't allow pressing the same button many times

### LINKED ISSUE #1949: It's possible to add multiple elements from one 'add element view'
Steps to reproduce
* Tap Add Element in the Page Editor
* Quickly tap the image icon as many times as you can before the Add Element view closes
* Check how many toucans you added, I'm able to add up to 4

You should only be able to add one element at a time.

**Update**
If you add two elements in this way, you will only ever be able to select one of them. You can move the one you can't select, but you can never delete or edit it.

errietta:
Closes #1949

### DIFF
--- a/www_src/pages/page/page.jsx
+++ b/www_src/pages/page/page.jsx
@@ -27,6 +27,7 @@ var Page = React.createClass({
       styles: {},
       currentElementId: -1,
       showAddMenu: false,
+      disableButtons: false,
       dims: {
         width: 0,
         height: 0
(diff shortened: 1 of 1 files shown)


www_src/pages/page/page.jsx
```diff
@@ -164,6 +166,10 @@ var Page = React.createClass({
           state.currentElementId = id;
         }
         this.setState(state);
+
+        setTimeout(function() {
```
thisandagain: Why the 1000ms timeout?

errietta: I had an issue with the menu taking a few ms to hide, during which you could again click the buttons

thisandagain: Gotcha. In that case, I'd recommend we match it to the transition time in the CSS (`0.2s`) and add a comment that explains why JIK that animation value ever gets changed.

thisandagain:
R+!

www_src/pages/page/page.jsx
```diff
@@ -164,6 +166,14 @@ var Page = React.createClass({
           state.currentElementId = id;
         }
         this.setState(state);
+
+        //Ensure we don't reach a race conditions where the buttons are
+        //re-enabled before the menu is hidden. 200ms matches the css animation
+        //speed.
+
+        setTimeout(function() {
```
Pomax: this should not have been here. This should not be a timeout, but should have been triggered by the child element being done doing what it's doing, informing its parent that it's now done and the buttons can be shown again. In traditional HTML+JS you would use a timeout, but in React, using a timeout is a sign that lifecycles are not being respected.

thisandagain: Can you please provide some guidance / links / resources for @errietta to take a look at?

Pomax: the best place to start is to work through the "getting started" and "tutorial" sections on http://facebook.github.io/react, to get an understanding of how React wants you to structure things, and how to make use of the life cycle functions to trigger things. After that it depends a little on how much you picked up, but I'd be happy to help find further tutorials/resources on how to do specific things the "Reacty" way after running through React's own tutorial.

PR #2058, opened 2015-06-02
setting link destinations and returning to source page

gvn:


gvn:
I'm gonna add some stuff and re-open...

PR #2059, opened 2015-06-02
save elements in the editing view when the back button is pressed

### LINKED ISSUE #2040: Element - Updates to elements are often not reflected when returning to Page view
STR:
https://www.dropbox.com/s/vapc2kysvoid36w/bug.mp4?dl=0
- Updates to element not reflected when returning to page view (0:14)
- Returning to the element editor and then going back will often resolve the issue (0:18 - 0:22)
- Two working attempts afterwards (inconsistent behavior)

/cc @secretrobotron @LauraReynal

Pomax:
fixes #2040 by initiating element save when the back button is pressed, rather than when componentDidUpdate triggers (which does not trigger on first time edit).

Test STR:
* create new project
* edit the empty page
* add a text element to the page
* edit the text element
* in the editor, change the text color to something
* hit the back button

In the current code, this will not update the element on the page, the edit appears "lost". With this PR, the element will update properly by saving changes (if there are any changes) to the element any time the back button is pressed, rather than hooking into React lifecycle functions.

### DIFF
--- a/www_src/pages/element/element.jsx
+++ b/www_src/pages/element/element.jsx
@@ -35,7 +35,7 @@ render(React.createClass({
 
     var saveBeforeSwitch = function() {
       var goBack = function() {
-        window.Android.goBack()
+        window.Android.goBack();
       };
       if (!this.edits) {
         return goBack();
--- a/www_src/pages/element/element.jsx
+++ b/www_src/pages/element/element.jsx
@@ -32,23 +32,34 @@ render(React.createClass({
 
   componentWillMount: function() {
     this.load();
+
+    var saveBeforeSwitch = function() {
+      var goBack = function() {
+        window.Android.goBack()
+      };
+      if (!this.edits) {
+        return goBack();
+      }
+      this.save(goBack);
+    }.bind(this);
+
+    this.props.update({
+      onBackPressed: saveBeforeSwitch
+    });
   },
   componentDidUpdate: function (prevProps) {
     // resume
     if (this.props.isVisible && !prevProps.isVisible) {
       this.load();
     }
-    // pause - if there are edits
-    if (this.edits && !this.props.isVisible && prevProps.isVisible) {
-      this.save();
-    }
   },
   cacheEdits: function (edits) {
     this.edits = edits;
   },
-  save: function () {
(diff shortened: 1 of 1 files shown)


www_src/pages/element/element.jsx
```diff
@@ -32,23 +32,34 @@ render(React.createClass({
...
+
+    var saveBeforeSwitch = function() {
+      var goBack = function() {
+        window.Android.goBack();
+      };
+      if (!this.edits) {
+        return goBack();
+      }
+      this.save(goBack);
+    }.bind(this);
```
k88hudson: if you use an arrow function you don't need this

www_src/pages/element/element.jsx
```diff
@@ -32,23 +32,34 @@ render(React.createClass({
...
+
+    this.props.update({
+      onBackPressed: saveBeforeSwitch
```
k88hudson: what do you think about doing something like this?
```js
onBackPressed: () => {
  if (!this.edits) {
    window.Android.goBack();
  } else {
    this.save(()=> window.Android.goBack());
  }
}
```

www_src/pages/element/element.jsx
```diff
@@ -32,23 +32,34 @@ render(React.createClass({
...
-    // pause - if there are edits
-    if (this.edits && !this.props.isVisible && prevProps.isVisible) {
-      this.save();
-    }
   },
   cacheEdits: function (edits) {
     this.edits = edits;
   },
-  save: function () {
+  save: function (postSave) {
```
k88hudson: might be clearer to call this `callback`

k88hudson:
@Pomax awesome, works well. Could you add a loading state in as well using the `Loading` component?

www_src/pages/element/element.jsx
Pomax: I kind of hold the completely opposite position: calling things `callback` is the worst naming, because it doesn't tell us anything about the variable. I'd be happy to prefix an `on` so that it's obvious that this is a callback function, though, while preserving the immediate information on what it will be used for.

k88hudson:
Since this is blocking, filing the loading state as a follow up ticket.

PR #2064, opened 2015-06-02
set and follow link destinations from editing views

gvn:


gvn:
Still some network loading issues...sometimes links don't update fast enough when returning to Page view.

gvn:
Related to #1670

www_src/pages/project/project.jsx
```diff
@@ -437,6 +437,30 @@ var Project = React.createClass({
...
   },
 
+  setDestination: function () {
+    // var pageUrl = `/projects/${this.state.params.project}/pages/${this.state.routeData.pageID}`;
+
+    api({
+      method: 'patch',
+      uri: `/users/1/projects/${this.state.params.project}/pages/${this.state.routeData.pageID}/elements/${this.state.routeData.elementID}`,
+      json: {
+        attributes: {
```
gvn: This is blowing away CSS changes if any are made before a destination is set.

gvn:
One other thing I've noticed: If you change the button color, then set link dest, the color change gets lost.

k88hudson:
@gvn might be a good idea to call `save` before switching the view to choose the link destination in that case then

www_src/pages/element/link-editor.jsx
```diff
@@ -28,8 +27,18 @@ var LinkEditor = React.createClass({
     }
 
   },
-  onChangeLinkClick: function () {
-    this.refs.notImplementedWarning.show();
+  onDestClick: function () {
+    var metadata = {
+      elementID: this.props.params.element,
+      pageID: this.props.params.page
```
k88hudson: i think you need to pass `projectId` as well -- right now, it doesn't matter since links only link to the same page, but in the future this could cause bugs

www_src/pages/project/project.jsx
```diff
@@ -212,7 +213,33 @@ var Project = React.createClass({
       }
 
     });
+
+    // Handle button actions on zoomed in pages
+    dispatcher.on('linkClicked', function (event) {
```
k88hudson: `(event) => {` removes the need for `bind`

www_src/pages/project/project.jsx
k88hudson: yeah, unfortunately you're going to have to pass along the rest of the attributes with `routeData`. Alternatively you could pass the chosen page *back* to the editor in `routeData`, instead of just triggering `goBack`, but i think the former solution would be a lot easier.

k88hudson:
A few small comments but this is looking good!

www_src/pages/project/project.jsx
gvn: Is that auto-binding a React thing or an ES6 thing?

k88hudson: it's an ES6 thing; arrow functions bind to the parent context, so they inherit both `this` and `arguments`

gvn: Nice!

www_src/pages/project/project.jsx
gvn: I wonder if it's possible to update the API to accept JSON "patches" instead of blowing away the entire object every time? @cadecairos any idea?

cadecairos: The reason I set it up the way it is is because if I merge the incoming JSON with the existing JSON in the db, there's no simple way to delete existing attributes besides setting them to `''` or `undefined`

gvn:
@k88hudson - Ok, everything should be working now!

www_src/components/el/types/link.jsx
```diff
@@ -47,6 +58,14 @@ var Link = React.createClass({
     });
   },
 
+  onClick: function () {
+    if (this.state.editing) {
+      this.activate();
+    } else {
+      dispatcher.fire('linkClicked', this.props);
```
k88hudson: can you remind me what this is for again? /me forgets

gvn: It allows this event's actions to be decoupled...basically I put the business/UI logic at a higher level.

www_src/pages/project/project.jsx
```diff
@@ -437,6 +468,32 @@ var Project = React.createClass({
...
+
+    patchedState = types.link.spec.expand(patchedState);
+
+    // Patch old attributes object to prevent overwritten properties
+    patchedState.attributes.targetPageId = this.state.selectedEl;
+    patchedState.attributes.targetProjectId = this.state.params.project;
+
+    api({
+      method: 'patch',
+      uri: `/users/1/projects/${this.state.params.project}/pages/${this.state.routeData.pageID}/elements/${this.state.routeData.elementID}`,
```
k88hudson: This will break when we change links to be able to refer to different projects, do you mind passing along the projectId in the routeData?

k88hudson:
This feature is ballin',  R+!

thisandagain:
:soccer: :basketball: :baseball: :football:

gvn:
rmypuj6

PR #2067, opened 2015-06-02
Initial zindex work

Pomax:
Addresses part of https://github.com/mozilla/webmaker-android/issues/2019

- refactors element-group.jsx for faster development by splitting up the large render() call into a set of smaller, easier to maintain functions
- gives page.jsx a knowlege of what the highest currently used z-index is when creating new elements, so that it can assign them a z-index that is 1 higher, ensuring new content is always on top of old content.

This patch does **not** add controls for raising/lowering element z-indices. This will be done as a follow up

### DIFF
--- a/www_src/components/element-group/element-group.jsx
+++ b/www_src/components/element-group/element-group.jsx
@@ -10,6 +10,12 @@ var ElementGroup = React.createClass({
       interactive: false
     };
   },
+  /**
+   * Generate the JSX for the element
+   * @param {id} elementId the element's id in the this.props-passed `elements` dictionary
+   * @param {object} elProps The element property sheet from which to build the JSX representation
+   * @return {JSX} an element's JSX representation
+   */
   formElement: function(elementId, elProps) {
     return (
       <div className={'el-wrapper' + (elProps.isCurrent ? ' current' : '')}>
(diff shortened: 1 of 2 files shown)


www_src/pages/page/page.jsx
```diff
@@ -147,9 +149,18 @@ var Page = React.createClass({
     });
   },
 
+  getHighestIndex: function() {
+    return Object.keys(this.state.elements)
+                 .map(e => this.state.elements[e].zIndex)
+                 .reduce((a,b) => a > b ? a : b, 1);
```
secretrobotron: :clap: :clap: :clap: 
so interesting to read es6 code

www_src/components/element-group/element-group.jsx
```diff
@@ -10,39 +10,44 @@ var ElementGroup = React.createClass({
       interactive: false
     };
   },
-  render: function () {
-    var elements = this.props.elements;
+  formElement: function(elementId, elProps) {
```
thisandagain: I'd love to see some documentation for these new methods :smile:

thisandagain:
Great. This looks good and is working on device. I'd love to see docs for each non-react method, but aside from that R+.

PR #2068, opened 2015-06-02
Fix editors to receive subsequent updates

k88hudson:


www_src/pages/element/link-editor.jsx
```diff
@@ -13,8 +13,20 @@ var LinkEditor = React.createClass({
   getInitialState: function () {
     return LinkBlock.spec.flatten(this.props.element, {defaults: true});
   },
-  componentDidUpdate: function () {
-    this.props.cacheEdits(this.state);
+  componentDidUpdate: function (prevProps, prevState) {
+    var state = this.state;
+
+    // Update state if parent properties change
+    if (this.props !== prevProps) {
```
gvn: Isn't this comparing 2 objects?

k88hudson: it actually looks like React passes the reference this.props if prevProps hasn't been updated

k88hudson: which is very convenient!

PR #2069, opened 2015-06-02
followup to the zindex patch

Pomax:
(built on top of the zindex patch, will rebase once that lands)

assigns an element the highest zindex when it is clicked/tapped without positional changes.

Note that this does not collapse the zindexing, it simply moves an element to "1 higher than highest". This might lead to large numbers, but even with 1000 elements on the page, a user would have to reorder things 2 million times, which at a second per reordering would take them over 23 days of nonstop clicking. Or, more realistically, about 45 years of spending 2 minutes on reordering every day.

A followup will be filed to optimize zindex ordering, but that does not need to land for beta 2/2.


Pomax:
superceded by https://github.com/mozilla/webmaker-android/pull/2075

Issue #1750, opened 2015-05-10
[MIGRATED] Implement font-family dropdown


flukeout:
* It's got too much left padding
* Doesn't have a little arrow indicator on the right side
* The dropdown options don't use the corresponding fonts to display the names
* The dropdown has radio buttons as the selection metaphor

Compare to spec below (and next page showing dropdown treatment and selected state)
https://projects.invisionapp.com/share/BK2M18Q28#/screens/72332476?maintainScrollPosition=false

thisandagain:
@vazquez Can you please drop in a reference PNG and link to sketch file?

vazquez:
### Design Documentation

##### Text Attributes Screen

###### Text Dropdown Sketch File

https://www.dropbox.com/s/8px5x6vx4o20psc/Text%20Dropdown.sketch?dl=0


###### Notes

- The Font Family selector acts as a [Material Design dropdown list](http://www.google.com/design/spec/components/menus.html#menus-behavior). It's pretty much the same as a normal dropdown except it looks just a little bit different.
- The user has the freedom to select the text to be bold AND italicized AND underlined.
- The user can only select one text alignment property. As an example, the text cannot be left aligned AND right aligned.
- Selected states for Text Style and Text Alignment is represented by a blue 'active state', as shown below.

![text options 2](https://cloud.githubusercontent.com/assets/1119821/7462955/79d643a6-f26a-11e4-8727-f9c3b53ae50c.png)

##### Text Options with Dropdown selected

- Every font shown here is accessible through Google Fonts.

![text options dropdown](https://cloud.githubusercontent.com/assets/1119821/7463148/d1d6d1aa-f26b-11e4-8dda-17e85a7e6f99.png)

Pomax:
This issue is going to need more work after beta, because we can't style option elements inside select dropdowns in WebView.

toolness:
<a href="https://github.com/mozilla/webmaker-core/issues/75"><img src="https://raw.githubusercontent.com/toolness/learning-issue-migrator/master/img/migrated.png"></a>
This issue has been moved to https://github.com/mozilla/webmaker-core/issues/75.
Please don't comment on it here.

PR #2076, opened 2015-06-03
Adds the Roboto, Bitter, and Pacfico fonts

Pomax:
Tries to address #1750, but can't.

This PR adds the three fonts we use as webfonts, and refactors the font `<select>` as a mixin for the text and image element to make use of (drying out the code a little).

Unfortunately, WebView doesn't allow CSS styling on `<option>` elements, unlike Firefox or Chrome, it looks like. So we can't get a nice preview of what the font looks like without moving to a "not a `<select>` element anymore" drop down menu.

### DIFF
--- a/www_src/html/index.html
+++ b/www_src/html/index.html
@@ -9,5 +9,6 @@
 <body>
   <div id="app"></div>
   <script src="{{ js_src }}"></script>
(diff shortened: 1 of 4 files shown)


thisandagain:
This is great / simple / straightforward. R+

Issue #2073, opened 2015-06-03
Add further GA tracking


secretrobotron:
### Events
- [ ] Login
- [x] Create new project
- [ ] Share project

### Screen Views
- [x] Project by ID

adamlofting:
By events, we mean these: https://developers.google.com/analytics/devguides/collection/android/v4/events

thisandagain:
We've got screen views setup, but we can add events in the next heartbeat.

thisandagain:
To track events we'll need a way to access this within JS. The Java API looks a little like this:
```java
// Get tracker.
Tracker t = ((AnalyticsSampleApp) getActivity().getApplication()).getTracker(
    TrackerName.APP_TRACKER);
// Build and send an Event.
t.send(new HitBuilders.EventBuilder()
    .setCategory(getString(categoryId))
    .setAction(getString(actionId))
    .setLabel(getString(labelId))
    .build());
```

Pseudo-code for a JS interface:
```js
if (window.Android) {
    window.Android.trackEvent('category', 'action', 'label');
}
```

thisandagain:
@adamlofting Really simple JS API for tracking events the Android GA tracker has landed: https://github.com/mozilla/webmaker-android#google-analytics-event-firing

Just holler if you have any questions or concerns!

adamlofting:
@thisandagain @ryanw-se that GA interface is wonderful! :boom: 

I've just got setup with a dev environment this evening and submitted this PR for tracking the 'Create new Project': https://github.com/mozilla/webmaker-android/pull/2089

ryanw-se:
@thisandagain The Android label can probably be removed from this, since a JS API was landed and all that remains is for the events to be added on the JS side.

thisandagain:
@ryanw-se Good catch. Removed! :boom:

adamlofting:
Events for `login` and `share` should be covered here: https://github.com/mozilla/webmaker-android/pull/2159

PR #2079, opened 2015-06-04
Expose Google Analytics tracker methods to WebAppInterface

ryanw-se:
This exposes two methods both called ```trackEvent()```. When called it will send a event to Google Analytics with the provided category, action, label and optionally a integer value. This feature allows for events to be called on the JavaScript side of things and partially solves #2073.

/cc @thisandagain @adamlofting

### DIFF
--- a/app/src/main/java/org/mozilla/webmaker/activity/Element.java
+++ b/app/src/main/java/org/mozilla/webmaker/activity/Element.java
@@ -5,12 +5,12 @@ import android.graphics.Bitmap;
 import android.os.Environment;
 import android.util.Log;
 
+import java.io.File;
+import java.util.UUID;
+
 import org.mozilla.webmaker.R;
 import org.mozilla.webmaker.WebmakerActivity;
 import org.mozilla.webmaker.util.Image;
-
-import java.io.File;
-import java.util.UUID;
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/mozilla/webmaker/activity/Element.java
```diff
@@ -5,13 +5,13 @@
...
 import android.util.Log;
 
-import java.io.File;
-import java.util.UUID;
-
 import org.mozilla.webmaker.R;
 import org.mozilla.webmaker.WebmakerActivity;
 import org.mozilla.webmaker.util.Image;
 
+import java.io.File;
```
thisandagain: Is your IDE making all these style changes automatically? If so, can you please just commit the changes needed to land the feature?

thisandagain:
Excellent! If you could limit this to just your changes in line 183 - 213 in `WebAppInterface.java` we should be good to go. Could you also add some documentation to the Android section of `README.md`?

ryanw-se:
@thisandagain Thanks for pointing that out, I went ahead and reverted that as well as adding some documentation involving Google Analytics events for JS in the ```android.md``` file.

PR #2080, opened 2015-06-04
[WIP] do not merge - snackbar notification

Pomax:
currently test-implemented in the link-editor view, with a button that can be pressed to pop up the snackbar-style notification

### DIFF
--- a/www_src/components/snackbar/snackbar.jsx
+++ b/www_src/components/snackbar/snackbar.jsx
@@ -1,0 +1,31 @@
+var React = require('react/addons');
+
+var Snackbar = React.createClass({
+  getInitialState: function() {
+    return {
+      message: this.props.message,
+      hidden: true
+    };
+  },
+  render: function () {
+    var className = "snackbar";
+    if (this.state.hidden) {
+      className = "hidden " + className;
+    }
+    return <div className={className}>{ this.state.message }</div>;
+  },
+  showMessage: function(msg) {
+    this.setState({
+      message: msg,
+      hidden: false
+    }, function() {
+      setTimeout(() => {
+        this.setState({
+          hidden: true
+        });
+      }, 3000)
+    });
+  }
+});
+
+module.exports = Snackbar;
--- a/www_src/components/snackbar/snackbar.jsx
+++ b/www_src/components/snackbar/snackbar.jsx
@@ -14,17 +14,29 @@ var Snackbar = React.createClass({
     }
     return <div className={className}>{ this.state.message }</div>;
   },
+  /**
+   * Hide the snackbar from view
+   */
   hide: function() {
     this.setState({
       hidden: true
     });
   },
+  /**
+   * Public API: show a message using the snackbar notification
+   * system. This slides the message into view at the bottom of
+   * the screen following the "snackbar" pattern from Material
+   * Design, leaving the message up for 3 seconds before sliding
+   * it back out of view.
+   */
   showMessage: function(msg) {
     this.setState({
       message: msg,
       hidden: false
     }, function() {
-      // TODO: ideally this is an event, but CSS can't inform
+      // TODO: ideally this is an event, but the duration is going
+      //       to have to be hard coded *somewhere*, and it's probably
+      //       best to keep that here, rather than in LESS code.
       setTimeout(_ => this.hide(), 3000);
     });
   }
--- a/www_src/components/snackbar/snackbar.jsx
+++ b/www_src/components/snackbar/snackbar.jsx
@@ -14,16 +14,22 @@ var Snackbar = React.createClass({
     }
     return <div className={className}>{ this.state.message }</div>;
   },
+  rehide: function() {
+    this.setState({
+      hidden: true
+    });
+  },
   showMessage: function(msg) {
     this.setState({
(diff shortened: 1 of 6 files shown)


www_src/pages/element/element.jsx
```diff
@@ -81,17 +80,28 @@ render(React.createClass({
...
-
       this.setState({element: data.element});
-
     });
   },
+
+  testSnackbar: function() {
+    this.recordError("Oh dear someone pressed the button");
+  },
+
```
Pomax: The `recordError` function is probably going to get put in its own mixin, along with a `generateSnackbar()` function so that any component can include the mixin without needing to require separate components etc.

www_src/components/snackbar/snackbar.jsx
```diff
@@ -0,0 +1,31 @@
...
+      className = "hidden " + className;
+    }
+    return <div className={className}>{ this.state.message }</div>;
+  },
+  showMessage: function(msg) {
+    this.setState({
+      message: msg,
+      hidden: false
+    }, function() {
+      setTimeout(() => {
```
Pomax: ideally this isn't a timeout, but the "3 seconds" needs to be hard-coded somewhere, and it seems more sensible to do so on the JS side than putting it in the CSS and the listening for transition events to change the state.

www_src/pages/element/element.jsx
```diff
@@ -20,7 +20,10 @@ var testIds = {
 };
 
 render(React.createClass({
-  mixins: [router],
+  mixins: [
+    router,
```
thisandagain: Nitpick: Moving the `require` into the `mixins` array as per the new `snackbar-mixin` is a good idea. Would be nice to have consistency.

www_src/pages/element/element.jsx
```diff
@@ -20,7 +19,10 @@ var testIds = {
 };
 
 render(React.createClass({
-  mixins: [router],
+  mixins: [
+    require('../../lib/router.jsx'),
+    require('../../lib/snackbar-mixin')
```
Pomax: if we put this into `Base`, that should add snackbar notifications for all views

www_src/lib/snackbar-mixin.js
```diff
@@ -0,0 +1,56 @@
+var React = require("react/addons");
+var Snackbar = require('../components/snackbar/snackbar.jsx');
+
+module.exports = {
+  /**
```
Pomax: this still needs a `componentDidMount` function that sets up dispatcher listening for "webmaker:error" or the like, which can then trigger the `recordError` function, for elements that do not have direct reference access to the Snackbar element.

www_src/components/snackbar/snackbar.jsx
```diff
@@ -0,0 +1,45 @@
...
+      message: this.props.message,
+      hidden: true
+    };
+  },
+  render: function () {
+    var className = "snackbar";
+    if (this.state.hidden) {
+      className = "hidden " + className;
+    }
+    return <div className={className}>{ this.state.message }</div>;
```
gvn: Couldn't you just do:

`return <div className="snackbar" hidden={this.state.hidden}>{ this.state.message }</div>;`

www_src/components/snackbar/snackbar.less
```diff
@@ -0,0 +1,17 @@
...
+  width: 100%;
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  background: #444;
+  line-height: 3em;
+  text-align: center;
+  color: #EEE;
+  height: 3em;
```
gvn: Try not to set explicit heights on elements that contain live text.

This leads to lots of problems when strings get longer (localization...etc).

gvn:
Am I correct in assuming there is only 1 snackbar per app? If that's the case I'm not sure why we'd want to regenerate it for every component that uses it via a mixin (if I'm understanding that correctly). We could just have 1 global instance that gets invoked via the router like I'm doing w. the modal and spindicator.

www_src/components/snackbar/snackbar.jsx
Pomax: hidden doesn't animate nicely, while `height` does. The design docs dictate this has to slide out after 3 seconds.

www_src/components/snackbar/snackbar.less
Pomax: in this case, it kind of has to be, since we need to animate it out of existence. We can, however, use vh or something similar to match what Material Design requires use to use.

Pomax:
There's only one per app, I initially wrote it for individual views, but adding it to the `Base` does make more sense. It's still an excellent idea to keep it as a mixin, so we're not bloating the Base class, and it stays easily refactorable, but it'll be a mixin that doesn't need to be used by individual views.

www_src/components/snackbar/snackbar.less
gvn: I'd animate positioning in that case. You may need some JS to accomplish this, but I can pretty much guarantee it will break with fixed height.

k88hudson:
As discussed, @pomax is going to finish this up and land on Friday

PR #2082, opened 2015-06-04
Allow JS interfacing users to send GA events

ryanw-se:
This exposes two methods both called trackEvent(). When called it will send a event to Google Analytics with the provided category, action, label and optionally a integer value. This feature allows for events to be called on the JavaScript side of things and partially solves #2073.

/cc @thisandagain @adamlofting

This is a compressed version of the changes in PR #2079.


thisandagain:
:+1: R+

PR #2085, opened 2015-06-04
Add login

k88hudson:
In this patch:

* sign-up/sign-in UI (`Login` activity)
* Login view is launched if no user session is found
* validation on each field based on server validation via `validations.js` mixin
* `jsonUtils` for parsing/form encoding, with tests
* new methods to `api` module added to create accounts/sign in, with tests
* new `getUserSession` method on WebAppInterface and `router.js`, with tests
* log out button on profile/Make view

TODO
* a few design details on login view and profile view (the shrinking header, tabbing doesn't seem to work between fields)


www_src/lib/api.js
```diff
@@ -62,4 +65,127 @@ module.exports = function (options, callback) {
...
+}
+
+api.BASE_URI = 'https://webmaker-api.herokuapp.com';
+api.BASE_LOGIN_URI = 'https://id.mofostaging.net';
+
+api.AUTHENTICATE_URI = api.BASE_LOGIN_URI + '/login/oauth/access_token';
+api.SIGN_UP_URI = api.BASE_LOGIN_URI + '/create-user';
+api.USER_URI = api.BASE_LOGIN_URI + '/user';
+
+// #authenticate (object options, function callback)
```
thisandagain: We are using [DocBlockr](https://packagecontrol.io/packages/DocBlockr) for JavaDoc-style documentation elsewhere.

www_src/lib/jsonUtils.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+  jsonToFormEncoded: function (json) {
+    if (!json) {
+      return;
+    }
+    return Object.keys(json).map(key => {
+      return encodeURIComponent(key) + '=' + encodeURIComponent(json[key]);
+    }).join('&');
+  },
+  parseJSON: function (string) {
```
thisandagain: :+1:

www_src/pages/login/sign-up.jsx
```diff
@@ -0,0 +1,162 @@
...
+  //    3. redirects to /main
+  onSubmit: function (e) {
+    e.preventDefault();
+
+    var errors = this.getValidationErrors();
+    if (Object.keys(errors).length > 0) {
+      return;
+    }
+
+    var json = {
```
thisandagain: Could we call this `options`? I generally find `json` to be a little confusing for me whenever I run into it and it isn't a serialized string.

www_src/pages/make/make.jsx
```diff
@@ -3,23 +3,24 @@ var render = require('../../lib/render.jsx');
 var api = require('../../lib/api.js');
 var Card = require('../../components/card/card.jsx');
 var Loading = require('../../components/loading/loading.jsx');
+var router = require('../../lib/router');
 
 var Make = React.createClass({
-  mixins: [],
+  mixins: [router],
```
thisandagain: We've started to do:
```js
mixins: [
    require('../blah/blah')
]
```
in other components as to keep the module namespace a little more terse / clear.

thisandagain:
:dancer: :star: :dancer: :star: :dancer: :star: This is so exciting!

www_src/pages/project/project.jsx
```diff
@@ -37,7 +37,8 @@ var Project = React.createClass({
   },
 
   uri: function () {
-    return `/users/1/projects/${this.state.params.project}/pages`;
+    var user = this.state.user || {};
+    return `/users/${user.id}/projects/${this.state.params.project}/pages`;
```
thisandagain: This breaks viewing projects from the discover list as we should be using the route params for both user id and project id. We should add a check to force the mode into "play" if the route param user id !== the session user id.

www_src/lib/validators.js
```diff
@@ -0,0 +1,98 @@
...
+    upperCase: {
+      regex: /[A-Z]+/,
+      message: 'Must contain an upper case letter.'
+    },
+    numbers: {
+      regex: /\d+/,
+      message: 'Must contain a number.'
+    },
+    username: {
+      regex: /^[a-zA-Z0-9\-]{1,20}$/,
```
thisandagain: Does this match our rules from `id.webmaker.org`?

www_src/lib/validators.js
```diff
@@ -0,0 +1,98 @@
...
+    lowerCase: {
+      regex: /[a-z]+/,
+      message: 'Must contain a lower case letter.'
+    },
+    upperCase: {
+      regex: /[A-Z]+/,
+      message: 'Must contain an upper case letter.'
+    },
+    numbers: {
+      regex: /\d+/,
```
thisandagain: I assume this matches `id.webmaker.org`?

k88hudson: Rules are here (id.webmaker.org uses the default settings) https://github.com/cadecairos/PassTest

thisandagain:
Alright! Let's land this thing. R+ Want to squash your commits?

Issue #2065, opened 2015-06-02
Improve transitions between views 


xmatthewx:
Problem: there is a delay to updates when exiting one view and entering another. View loads between elements update, for example.  

Solution 1 of 2: engineering will improve timing and efficiency of update flow.

Solution 2 of 2: @vazquez will design a interstitial indeterminate loading state indicator  

cc @k88hudson

vazquez:
We need something simple and peppy. What about this? http://codepen.io/rvazquez/pen/vOxXJL

k88hudson:
Note that some UI needs to be added for this bug fix https://github.com/mozilla/webmaker-android/pull/2059

xmatthewx:
Looks great @vazquez. (that codepen works in FF only) 

@k88hudson - do you expect this to fire between every single view transition, or only if the network lags a bit? I'm concerned it will get annoying.

xmatthewx:
Add a slight delay before the overlay appears. We might earn a moment of patience with the user if we animate the arrival of it.

k88hudson:
Note that this style of loading should only be implemented when network events are triggered when back is pressed (on the element editor, settings view); it should not replace the less obtrusive loading state for more granular updates like adding/deleting pages etc.

gvn:
@xmatthewx I set the default delay to 1 second, which feels pretty good, but I can change it to whatever.

500ms felt a bit too short FWIW.

PR #2087, opened 2015-06-04
basic Spindicator

gvn:


gvn:
Resolves #2065

thisandagain:
Looks great and is working well when I set my device to 2G or 3G. `1000`ms feels like the right default.

www_src/pages/page/page.jsx
```diff
@@ -298,6 +298,7 @@ var Page = React.createClass({
 
   load: function() {
     api({
+      spinOnLag: false,
```
thisandagain: Why no spinner here?

thisandagain:
R+ :spaghetti: :sheep:

Issue #1569, opened 2015-04-27
Implement remix flow


xmatthewx:
depends on #1566

vazquez:
Build UI as specified in https://github.com/mozilla/webmaker-android/issues/1566#issuecomment-98890892

xmatthewx:
Note: 
 - All v1 users will be required to sign up / login. Design the sign-up prompt since we've already had copy review. But no need to implement now.  
 - Projects be names "Remix of Foo"
 - First-time remix prompt can be implemented to show every time until it proves to actually be annoying.

xmatthewx:
Revising previous note for clarity: 

 - All v1 users will be required to login, so let's just implement the dialog below.
 - Projects will be named "Remix of Foo..."
 - I wonder if we should show this dialog the first time (unobtrusive), or every time (clarity). 

![screenshot 2015-05-12 18 06 02](https://cloud.githubusercontent.com/assets/53411/7599624/b274461e-f8d1-11e4-8d73-b9174bc0e95e.png)

vazquez:
![remix](https://cloud.githubusercontent.com/assets/1119821/7634695/9da1edcc-fa2a-11e4-9aa9-5d7316b327ba.png)

vazquez:
Or this one:

![remix copy](https://cloud.githubusercontent.com/assets/1119821/7635044/9b824116-fa2c-11e4-91fe-0393a82f9ba0.png)

xmatthewx:
Revised copy:

> **Project Remix**
> This is your copy of **My Yam Fries Recipe**. You can add or change anything. The original by _fancyunicorn_ will stay the same. Have fun!

Your use of the hex icon prompted an idea: Remix is the tinker version of make. Let's match the tinker link dialog and icon with the phrase "Project Remix"
 
![screenshot 2015-05-14 13 13 03](https://cloud.githubusercontent.com/assets/53411/7637312/0d393888-fa3b-11e4-8fa6-e1be4ee948ec.png)

vazquez:
:+1: Will do

vazquez:
Updated treatment and copy.

![remix copy](https://cloud.githubusercontent.com/assets/1119821/7639468/2b46cb56-fa4a-11e4-84ec-c7ecce345404.png)

xmatthewx:
:+1: on UI and copy

Scenario: 
 1. user triggers remix from project header ellipses menu
 1. project is copied with the name "Remix of Foo bear"
[...]

PR #2092, opened 2015-06-05
modal window component

gvn:


gvn:
Progress on #1569 

This is just a static content version of the modal window. I'll be making this dynamic for broader usage.

You can launch it from the style guide "Show modal" button and hide it with the "ok got it" in the window.

www_src/components/modal/modal.jsx
```diff
@@ -0,0 +1,50 @@
+var React = require('react');
+var dispatcher = require('../../lib/dispatcher');
+
+var Modal = React.createClass({
+  getInitialState: function () {
+    return {
+      isVisible: false
```
Pomax: we're also using `isVisible` as a special state property for determining whether views are "in view" or frozen off screen into cache, so it might be safer to use a different term?

www_src/components/modal/modal.jsx
```diff
@@ -0,0 +1,50 @@
...
+      isVisible: true
+    });
+  },
+  hide: function () {
+    this.setState({
+      isVisible: false
+    });
+  },
+  componentDidMount: function () {
+    dispatcher.on('showModal', (event) => {
```
Pomax: I'm doing something similar for Snackbar notifications (oh god, Android terminology) over at https://github.com/mozilla/webmaker-android/pull/2080 that we might be able to combine (practice wise) for wins on both PRs: if we combine using the dispatcher from your PR with wrapping most of the functionality as a mixin from my PR, it feels like we'd have a powerful "Views that need to house the DOM element that actually achieves notification can do so, and everything that lives *in* a view can just dispatch modal and snackbar messages".

I would advise making the "showModal" a "modal:show" name though (because namespacing), and then handling the event with a separate function like
```
  componentDidMount: function () {
    dispatcher.on('showModal', this.showMessage);
    ...
  },

  showMessage: function(evt) {
    this.setState({
      message: evt.message,
      ...,
      isVisible: true
    });
  }
```

gvn: I am totally down for namespacing the message names. Good call!

I'm not sure I get the need for a separate function in this case. It seems extra verbose for no clear advantage. I'm already invoking local methods that double as an explicit API for the modal. With your approach the only way to trigger the visibility is through message passing via the dispatcher. With my approach you could just call `show` or `hide` directly from a parent via refs if you wanted to. The extra dispatcher dependency is only needed in non parent-child compositions.

www_src/components/modal/modal.jsx
gvn: Hmm. Any suggestions?

www_src/components/modal/modal.jsx
gvn: I'm also unclear on why you'd make `message` a property of state?

gvn: Actually...disregard that. I get it. It'll be needed once the modal's content isn't static... :+1:

www_src/components/modal/modal.jsx
```diff
@@ -0,0 +1,67 @@
...
+    }
+  },
+  componentDidMount: function () {
+    dispatcher.on('modal:show', (event) => {
+      this.setState(event.config);
+      this.show();
+    });
+
+    dispatcher.on('modal:hide', (event) => {
+      this.hide();
```
k88hudson: i think it would be a good idea to reset with `getInitialState` here

k88hudson: Otherwise subsequent modals will pick up the last model's config instead of defaults -- either that, or have assign(this.getInitialState, event.config) in the `modal:show` event

www_src/components/modal/modal.less
```diff
@@ -0,0 +1,70 @@
+#modal {
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background: rgba(0,0,0,0.3);
```
k88hudson: I would go for `fade(#1A2237, 0.7)`

k88hudson: (that's what the image modal background is -- although probably a good idea to save this as `overlay` or something in colors.less maybe?)

www_src/components/modal/modal.less
```diff
@@ -0,0 +1,70 @@
...
+  z-index: 2147483647; /* signed 32bit maxint */
+  display: flex;
+  padding: 30px;
+  justify-content: center;
+  align-items: center;
```
k88hudson: Could use a box-shadow as per the design

www_src/pages/style-guide/style-guide.jsx
```diff
@@ -34,6 +35,20 @@ var StyleGuide = React.createClass({
       this.setState({loading: false});
     }, 3000);
   },
+  showModal: function () {
+    dispatcher.fire('modal:show', {
+      config: {
+        header: 'Project Remix',
+        body: 'This is your copy of My Yam Fries Recipe. You can add or change anything. The original will stay the same. Have fun!',
+        attribution: 'fancyunicorn',
+        icon: '../../img/tinker.png',
```
k88hudson: Might be a good idea to add `../../img/` in the component, since this will help reduce path errors

k88hudson:
Looks great, a few suggestions

www_src/components/modal/modal.less
gvn: That becomes `rgba(26, 34, 55, 0.007)`, which is barely visible. Did you have a typo there?

www_src/components/modal/modal.less
gvn: :ballot_box_with_check:

www_src/components/modal/modal.less
k88hudson: oh whoops! i meant 70

Issue #2013, opened 2015-05-27
Hook up thumbnail to Discover view


xmatthewx:
Depends on https://github.com/mozilla/api.webmaker.org/issues/64

PR #2104, opened 2015-06-05
Hook up discover and make lists to '320' thumbnail

thisandagain:
Resolves GH-2013


thisandagain:
@Pomax R?

Pomax:
Hm, how can I test this? If I rebuild dev with this on top, and run the app, I do not see thumbnails in the discover tab.

Issue #2074, opened 2015-06-03
redo the button disabling by taking out the setTimeout


Pomax:
https://github.com/mozilla/webmaker-android/pull/2049 accidentally landed some code that makes sense in traditional HTML, but should never feature in React code. The button disabling should be immediate upon an addElement call, and the subsequent button enabling should be triggered by the element having been added, not via a timeout.

errietta:
@Pomax can has mentoring?

thisandagain:
Hey @Pomax, 
If I'm reading this correctly, it looks like @errietta is using the timeout not to wait for the element being added, but rather to wait for the CSS transition to complete. As otherwise the user could add another element in-between the time the element is added and the overlay / buttons transition off-screen (which takes 200ms). It looks like we could use the [ReactCSSTransitionGroup](https://facebook.github.io/react/docs/animation.html#high-level-api-reactcsstransitiongroup) to do what you are describing ... is that what you are thinking?

thisandagain:
Follow-up: @errietta is going to take another crack at it using the React animation APIs.

errietta:
sir @Pomax :3

PR #2105, opened 2015-06-05
Don't allow users to add multiple things from one button - no timeout

errietta:


errietta:
Closes #2074

www_src/pages/page/page.jsx
```diff
@@ -232,18 +244,7 @@ var Page = React.createClass({
...
-        //              to its "this.props.dosomething", which was passed down by
-        //              the parent, and then in this component that function would
-        //              lead to a this.setState({ disabledButtons: false }).
-        //
-        // https://github.com/mozilla/webmaker-android/issues/2074 has been filed to fix this
-
-        setTimeout(function() {
-          this.setState({disableButtons: false });
-        }.bind(this), 200);
+        this.setState({disableButtons: false });
```
Pomax: we can actually move this into the state that's getting set in L243 now, too.

www_src/pages/page/page.less
```diff
@@ -38,30 +38,41 @@
...
 
-  .overlay {
+  .overlay-enter {
     position: fixed;
     top: 0;
     left: 0;
     height: 100%;
     width: 100%;
-    z-index: 998;
+    z-index: 999;
```
Pomax: bizar as it seems, this is not high enough. Let's use 2147483647, the maxint value allowed for zindex, with a comment that explains WHY it's this high.

www_src/pages/page/page.less
```diff
@@ -38,30 +38,41 @@
...
+  /* See the end of https://facebook.github.io/react/docs/animation.html#getting-started */
+
   .controls {
     @_transformDistance: 55px;
     position: fixed;
     bottom: 20px;
     left: 0;
     right: 0;
-    z-index: 999;
+    z-index: 998;
```
Pomax: and then we can change this one to 2147483646, with a comment to explain it has to be "the highest - 1" so that the overlay still wins.

www_src/pages/page/page.jsx
```diff
@@ -92,7 +100,11 @@ var Page = React.createClass({
         </div>
       </div>
 
-      <div className={classNames({overlay: true, active: this.state.showAddMenu})} onClick={this.toggleAddMenu}/>
+      <div className="overlaydiv">
+        <ReactCSSTransitionGroup transitionName="overlay">
+          {placementOverlay}
```
Pomax: given that we've collapsed `placementOverlay` to just this to just `<div/>` now, let's take out the bit up top and change this line to the following instead:
```
{ this.state.disableButton ? <div/> : false }
```
and then this is totally an R+

www_src/pages/page/page.jsx
```diff
@@ -74,6 +76,8 @@ var Page = React.createClass({
       url = `/projects/${params.project}/pages/${params.page}/elements/${currentEl.id}/editor/${currentEl.type}`;
     }
 
+    // Overlay to show when elements are placing themselves on the page
```
Pomax: phantom comment

www_src/pages/page/page.jsx
```diff
@@ -92,7 +94,12 @@ var Page = React.createClass({
         </div>
       </div>
 
-      <div className={classNames({overlay: true, active: this.state.showAddMenu})} onClick={this.toggleAddMenu}/>
+      // Overlay to show when elements are placing themselves on the page
```
Pomax: React will treat this as literal text. Better to just not put it in, since the markup is fairly clear in what it's for.

Pomax:
R+! Although I'd like @vazquez to give the fade a look over to make sure we're keeping with the feel we want for the app.

www_src/pages/page/page.less
```diff
@@ -38,30 +38,44 @@
...
+    &.overlay-enter-active {
       visibility: visible;
       opacity: 0.7;
     }
   }
 
+  /* Necessary to force-remove ReactCSSTransitionGroup elements in unminified React: */
+  .overlay-leave {
+    opacity: 1;
+    transition: opacity .5s ease-in;
```
vazquez: Let's make this transition:

`transition: all 0.3s cubic-bezier(0.79, 0.14, 0.15, 0.86) 0s;`

thisandagain:
R+ !

Issue #1729, opened 2015-05-09
Hook up tinker editor


k88hudson:
With colour picker, real data updating, api, etc.

thisandagain:
:+1:

thisandagain:
Some initial work done here: https://github.com/thisandagain/webmaker-android/tree/gh1729

k88hudson:
Resolved!

PR #2115, opened 2015-06-08
#1729 - reinstate tinker mode

k88hudson:
Reimplement Tinker mode

* adds `getElement` and `updateElement` to api.js, with tests
* adds HSB, RGB, and Alpha "private" components to ColorSpectrum
* renames `attribute` param in router with `propertyName`, since tinker mode can be applied to both styles and attributes
* adds an `editor` property to element specs, so individual styles/attributes can specify a tinker mode editor
* a few updates to image-editor and text-editor to deal with updating issues


www_src/components/color-group/color-group.jsx
```diff
@@ -9,9 +9,8 @@ var ColorGroup = React.createClass({
   mixins: [React.addons.LinkedStateMixin],
   getDefaultProps: function () {
     return {
-      colors: this.defaultColors,
-      id: 'value',
-      tinkerUrl: '/users/1/projects/123/pages/0/elements/1/attributes/color/editor/color'
+      colors: this.defaultColors.slice(0),
```
Pomax: no need for `0` thanks to JS magic. If you want a copy `things.slice()` works great.

www_src/components/color-spectrum/color-spectrum.jsx
```diff
@@ -1,20 +1,116 @@
 var React = require('react');
+var Slider = require('../../components/range/range.jsx');
+var Color = require('color');
 
-// Doesn't do much yet, but renders the hue and saturation spectrum
+var HSB = React.createClass({
+  getBackgroundColor: function () {
+    var color = this.props.color.clone();
+    color.hsv(color.hue(), 100, 100);
```
Pomax: this is white, no matter what hue you pass in, because the lightness is set to max. Is that intentional?

www_src/components/color-spectrum/color-spectrum.jsx
```diff
@@ -1,20 +1,116 @@
...
+    this.props.onChange(color);
+  },
+  changeSB: function (e) {
+    var color = this.props.color;
 
-var ColorSpectrum = React.createClass({
+    var x = e.touches[0].clientX - this.bounding.left;
+    var y = e.touches[0].clientY - this.bounding.top;
+
+    color.hsv(color.hue(), x / this.bounding.width * 100, 100 - y / this.bounding.height * 100);
```
Pomax: for clarity, `100 * x / this.bounding.width` and `100 * (100 - y) / this.bounding.height` are safer, so your intent is clear, and people won't wonder about whether that should be `x / (this.bounding.width * 100)` or something

www_src/components/color-spectrum/color-spectrum.jsx
```diff
@@ -1,20 +1,116 @@
...
+  },
+  onTouchMove: function (e) {
+    this.changeSB(e);
+  },
   render: function () {
+
     return (<div className="spectrum-container">
       <div className="spectrum-left">
-        <div className="saturation" style={{backgroundColor: this.props.value}}>
+        <div ref="sl" className="saturation" style={{backgroundColor: this.getBackgroundColor()}} onTouchStart={this.onTouchStart} onTouchMove={this.onTouchMove}>
```
Pomax: strongly advice to just be verbose here. "sl" is not a very meaningful word, do have have a better candidate?

www_src/components/color-spectrum/color-spectrum.jsx
```diff
@@ -1,20 +1,116 @@
...
           <div className="saturation-white" />
           <div className="saturation-black" />
+          <div ref="marker" className="sl-marker" style={this.getSBPosition()}/>
         </div>
-        <input className="hue" type="range" orient="vertical" />
+        <input value={this.props.color.hue()} min={0} max={359} className="hue" type="range" orient="vertical" onChange={this.changeHue} />
```
Pomax: I wish we had `<input type="jogdial"...>` or something, since hue is circular =P

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: because of the way the colour spectrum is supposed to look, this is  `hsv/b`, not `hsl`

www_src/components/color-spectrum/color-spectrum.jsx
```diff
@@ -1,20 +1,116 @@
...
+    };
+  },
+  render: function () {
+    return (<div>
+      {['Red', 'Green', 'Blue'].map((label, i) => {
+        var lowercaseLabel = label.toLowerCase();
+        var currentValue = this.props.color[lowercaseLabel]();
+        return (<div key={label} className="form-group">
+          <label>{label}</label>
+          <Slider value={currentValue} max={255} onChange={this.generateOnChangeCallback(lowercaseLabel)} />
```
Pomax: `min={0} step={1}`?

www_src/components/tabs/tabs.jsx
```diff
@@ -31,7 +31,7 @@ var Tabs = React.createClass({
             'tab-body': true,
             'selected': this.state.selectedIndex === i
           });
-          return (<div className={className}>{tab.body}</div>);
+          return (<div key={i} className={className}>{tab.body}</div>);
```
Pomax: generally a bad idea. While it does stop React from complaining, there is no guarantee that the ordering for the array is stable. Is there a `tab.name` or something we can use instead so that React can do its diffing properly?

www_src/lib/api.js
```diff
@@ -86,6 +86,39 @@ api.AUTHENTICATE_URI = api.BASE_LOGIN_URI + '/login/oauth/access_token';
 api.SIGN_UP_URI = api.BASE_LOGIN_URI + '/create-user';
 api.USER_URI = api.BASE_LOGIN_URI + '/user';
 
+api.validateOptions = function (name, requiredFields, options) {
+
+  requiredFields = requiredFields || [];
+
+  if (typeof name !== 'string' || requiredFields instanceof Array === false) {
```
Pomax: why `instanceof` instead of testing for what it needs to support, like foreach or slice?

www_src/components/tabs/tabs.jsx
k88hudson: hmm... i might be able to use the icon, but it's not much more useful than `i`

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: those are the defaults

www_src/lib/api.js
```diff
@@ -200,4 +233,39 @@ api.signUp = function (options, callback) {
...
 
+/**
+ * [getElement Retrieves an element from the db]
+ * @param  {Object}   options  Should include ids for: user, project, page, and element
+ * @param  {Function} callback Called with two params, err and data,
+ *                             where err.message is a descriptive error message if an error occured,
+ *                             or data is an object representing the element.
+ */
+api.getElement = function (options, callback) {
+  api.validateOptions('getElement', ['user', 'project', 'page', 'element'], options);
```
Pomax: this array feels a bit magical. Can we refactor these to a mapping that can be maintained without needing to go into functions? Like
```
var requirements = {
  'getElement':  ['user', 'project', 'page', 'element'],
  'updateElement': ['user', 'project', 'page', 'element', 'json'],
  ...
};
```
and then refer to that here, like
```
  api.validateOptions('getElement', requirements['getElement'], options);
```
or just
```
  api.validateOptions('getElement', options);
```

with the validateOptions function pulling the necessary requirements?

www_src/pages/element/element.jsx
```diff
@@ -109,7 +109,7 @@ render(React.createClass({
 
     return (<div>
       <Editor {...props} />
-      <button hidden={window.Android} onClick={()=>this.save()}>SAVE</button>
+      <button hidden={window.Android} onClick={()=>this.save()}>DEBUG:SAVE</button>
```
Pomax: This can just be `onClick={this.save}`

www_src/pages/tinker/tinker.jsx
```diff
@@ -1,68 +1,157 @@
...
+      loading: false,
+      element: null
     };
   },
-  onChangeColor: function (color) {
+
+  componentDidMount: function () {
+
+    // Prevent pull to refresh
+    document.body.style.overflowY = 'hidden';
```
Pomax: can we take out this hack and make that a class that we can use instead? If it's style, let's not JS hardcode it - we'll never find it again =(

www_src/pages/tinker/tinker.jsx
```diff
@@ -1,68 +1,157 @@
...
+
+    this.load();
+
+    this.props.update({
+      onBackPressed: () => {
+        this.save(() => window.Android.goBack());
```
Pomax: <3 ES6, but: `this.save(window.Android.goBack)` will work just fine

www_src/pages/tinker/tinker.jsx
```diff
@@ -1,68 +1,157 @@
...
+    updateObject[isStyleOrAttribute] = spec.expand(element)[isStyleOrAttribute];
+
+    this.setState({loading: true});
+    var options = assign({}, this.state.params, {json: updateObject});
+    api.updateElement(options, (err, element) => {
+      this.setState({loading: false});
+      if (err) {
+        console.error(err);
+      }
+      if (typeof onSaveComplete === 'function') {
```
Pomax: do we need this check? Can we not just say `if (onSaveComplete) { onSaveComplete(); }`? If that throws an error, that's a good thing?

www_src/pages/element/element.jsx
k88hudson: at the moment the save function will try to call `e` param as a function, although we could add a typeof check

www_src/pages/tinker/tinker.jsx
```diff
@@ -1,68 +1,157 @@
...
+  getEditorValue: function () {
+    return this.state.element && this.state.element[this.state.params.propertyName];
+  },
+
+  setEditorValue: function (value) {
+    var edits = {};
+    edits[this.state.params.propertyName] = value;
     this.setState({
-      color: color
+      element: assign(this.state.element, edits)
```
Pomax: is this safe, or should this be `var thing = assign({}, this.state.element)` + `assign(thing, edits)` instead, so that we don't modify `this.state.element`?

www_src/pages/tinker/tinker.jsx
k88hudson: it's either that or we need to call `this.save` wrapped in an anonymous function when it's used as a listener callback

k88hudson: personally i prefer this, since it's less prone too accidentally forgetting to wrap this.save

Pomax:
We can reduce gray_and_white_checkers.gif considerably, since it's a repeating pattern - we just need the four tile pattern, and CSS repeat does the rest

www_src/pages/tinker/tinker.jsx
k88hudson: functionally it is safe, since it's in the middle of a setState call, but probably good practice to add {}

www_src/pages/tinker/tinker.jsx
k88hudson: personally i think this would be better served by an API on the root component, but i'd rather not refactor in this ticket (I used the same strategy as is used elsewhere so it would be easily find-and-replaceable), mind if I file as a follow-up?

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: oops! slipped by my find-replace :stuck_out_tongue_winking_eye:

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: Ok, does
```
x / this.bounding.width * 100, 100 - (y / this.bounding.height * 100))
```
look clear enough?

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: :red_circle: :blue_circle: :P

www_src/lib/api.js
k88hudson: hm, good idea!

www_src/pages/tinker/tinker.jsx
k88hudson: Unfortunately not doing this causes a Java error, although I have to admit I don't exactly know why

www_src/components/color-spectrum/color-spectrum.jsx
Pomax: ahhh, I said nothing =P

www_src/components/color-spectrum/color-spectrum.jsx
Pomax: not really, the multiplication after the division throws me off. I can't tell if you really mean `x / this.bounding.width * 100` or whether you meant `x / (this.bounding.width * 100)`, which is made apparent my moving the multiplication up front: `100 * x / this.bounding.width`. (Slightly better performance that way too, since you're not upscaling a lossy floating point, but simply dividing a safe integer)

www_src/components/color-spectrum/color-spectrum.jsx
Pomax: in part "I don't trust browsers to respect defaults", but mostly it's a matter of "let's not leave implicit what we can be explicit about,so that other devs/contributors reading the code immediately get it"

www_src/components/tabs/tabs.jsx
Pomax: Hmm... if the icon is tied to the tab, that's better than `i`, probably. That way even if the tabs array is reordered, React will still correctly identify each tab as "the one it should be".

www_src/pages/tinker/tinker.jsx
Pomax: as long as we're tracking it I have no problems with a follow up

www_src/pages/tinker/tinker.jsx
Pomax: that... sounds blog worthy O_o

Pomax:
Does this bring up any new UI to test with, or is this a matter of "the same UI should do the right thing still, despite the plumbing being different now"?

www_src/components/color-spectrum/color-spectrum.jsx
k88hudson: in this case these are `defaultProps` for the `Range` component; if you're disagreeing with the notion of having default props altogether, I'm going to have to disagree with you there

www_src/lib/api.js
k88hudson: In this case I specifically want whoever is using this function to supply an Array, because the data should be of that form.

k88hudson:
@Pomax tinker mode should now show up next to any color selector in the element editors, and it should actually work

www_src/pages/tinker/tinker.jsx
k88hudson: filed https://github.com/mozilla/webmaker-android/issues/2122

Pomax:
A few UI notes, but functionally this is great and I don't see any errors while running it.
- Tinker mode's a lot darker than everything else
- The heading still reads "Options", which I think we discussed as needing a change before we backburner'd tinker mode
- The "pencil" vs. "slidythings" tabs seem a bit of a departure from the rest of our UI, and don't look like tab context switchers (I couldn't figure out what they were until I tried tapping everything).
- The manipulable values have their numeric value next to them in green, rather than white, is that intentional?
- Going back from tinker mode in "slider" view to the regular editor, then going back to tinker mode shows tinker mode as "pencil" view again. It make might sense to remember which was picked for the element that's "currently being edited".

R+, provided we have a place to follow up on those questions

k88hudson:
@Pomax the colour is implemented according to https://projects.invisionapp.com/share/P62Z2HBMV, but feel free to file issues as a new ticket tagged "design"

PR #2124, opened 2015-06-09
Add support for external links

k88hudson:
This will resolve https://github.com/mozilla/webmaker-android/issues/2111, but reintroduces a problem we had before:

The issue is that because the `Login` activity is set to `noHistory="true"` in the manifest, tapping on an external link,which launches the browser, and then pressing the back button sends you to the screen *before* login.


k88hudson:
@thisandagain R? [[flagged!]](http://plzkthx.herokuapp.com)

thisandagain:
This is working great. :+1: R+

Issue #2113, opened 2015-06-08
Add "remix" to project dropdown


gvn:
Per #1569 we'll need "Remix" added to the ellipses dropdown for projects.

gvn:
@thisandagain Can you take this?

I think it will just need to redirect to the Project edit view with an additional param that I can look for to show the modal window. Something like `isFreshRemix` that is only truthy the first time around...

gvn:
I guess it will also need to trigger the copying and renaming of the target project as well...

xmatthewx:
With this, let's add More Actions menu to Play view. 

Includes:
  - Remix
  - Share
  - Edit (author only)

thisandagain:
I can handle the remix and share, but "edit" is going to need some extra work. Can you file another ticket @xmatthewx ?

xmatthewx:
"edit" filed in #2128

PR #2125, opened 2015-06-09
Add remix and sharing controls to play activity. Resolves GH-2113

thisandagain:
Selecting "Remix" in the `Play` activity triggers:
```java
view.load("javascript: window.createRemix && window.createRemix()", null);
```


gvn:
Call is coming through on the webview, can't comment too much on the Android stuff other than WFM! :cat2:

PR #2130, opened 2015-06-09
Clean up navigation styles on Android 4.2.X

thisandagain:
This adds some additional styles / clean-up for older versions of Android


thisandagain:
@k88hudson R?

k88hudson:
Nice! Is there an easy way to change the options menu to use a light theme instead of black?

thisandagain:
Not an easy way ... but we have a way. That is next on my list along with some type adjustments.

Issue #2131, opened 2015-06-09
Don't cache RouteData


gvn:
When passing metadata via `Android.setView` I'm expecting it to only be passed through to the target view initially. 

It seems like we're caching it, which is causing some problems...

gvn:
Blocker for #1569

PR #2132, opened 2015-06-09
Destroy route data after access from MemStorage. Resolves GH-2131

thisandagain:


thisandagain:
@gvn R?

gvn:
So, this appears to be preventing the data from ever being passed. I don't see it coming through at all with the patch... :sob:

www_src/lib/router.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
 
     if (this.android) {
       data = parseJSON(this.android.getRouteData());
+      this.android.clearRouteData();
```
gvn: I think this is always clearing the incoming route data...

PR #2135, opened 2015-06-10
fall back to endmark after second touch end has been registered

Pomax:
Fixes https://github.com/mozilla/webmaker-android/issues/2133 near as I can tell.

Added debugging that is off by default because I suspect this is not the last time we need to touch this file.


thisandagain:
Resolves the issue on Nexus 5 and Galaxy S. R+

PR #2136, opened 2015-06-10
zindex fixing, and refactors that made that even a tractible proposition

Pomax:
This fixes https://github.com/mozilla/webmaker-android/issues/2101, by ensuring we extract the "modified" flag **before** resetting the transform object.

This also includes a fair bit of refactoring that was necessary to get the code to a point where it was possible to read without getting too lost.


thisandagain:
Getting closer. I'm still able to see some confusing behavior after taping on an element to move it to the top z-index and then immediately dragging it. Screen recording here:
https://www.dropbox.com/s/f9xrtzh9z37t8s8/gh2101.mp4?dl=0

/cc @Pomax @xmatthewx

Pomax:
The lack of seeing when/where you're tapping is making this video not as useful as hoped to see what might still be going on. Can you rerecord with visual taps, or text STR?

thisandagain:
STR:
- Add three elements to a page
- Scale and position each
- Position one so that it overlaps another and you can still "tap" on both
- Tap to move one to front
- Drag that element

Expected:
- Element maintains current `z-index` while dragging

Actual:
- Element seems to go back to it's original `z-index` while dragging

Pomax:
confirmed. I can make this happen with two elements as well, so let's find out wtf is still going wrong.

Pomax:
This now comes with some heavy refactoring that was necessary to track when and where things happened (imagine this code having console.logs at every function entry and code branch), so definitely need more than one pair of eyes on it. @thisandagain, @k88hudson, can I ask both of you to review?

www_src/components/basic-element/basic-element.jsx
```diff
@@ -78,31 +85,46 @@ var El = React.createClass({
...
+    //
+    // TODO: this discovery might mean we need to do this for all
+    //       transform properties too, but there haven't been any
+    //       bugs filed about those (yet...?)
+    if(this.props.zIndex !== this.state.zIndex) {
+      this.setState({
+        zIndex: this.props.zIndex
+      });
+    }
+  },
```
Pomax: This is the actual fix. Everything else is refactors and cleanup to get to this point, and while it's tempting to say "let's refactor separately", I **strongly** urge us not to. Let's take this refactor as necessary for this fix to even exist.

www_src/pages/page/page.jsx
```diff
@@ -35,14 +31,29 @@ var Page = React.createClass({
   },
 
   componentWillMount: function() {
+    // who are we?
+    var params = this.state.params;
+    this.uri =  `/users/${params.user}/projects/${params.project}/pages/${params.page}`;
```
k88hudson: this will break if you change users/change projects etc. without fully reloading the page

www_src/components/basic-element/basic-element.jsx
```diff
@@ -40,11 +40,11 @@ var El = React.createClass({
       zIndex: this.props.zIndex,
       touchactive: false
     };
+    this.buttonStyle = initial;
```
k88hudson: What is this for?

www_src/pages/page/page.jsx
Pomax: yep, can do

www_src/components/basic-element/basic-element.jsx
Pomax: this was `buttonStyle = this.getInitialState()`, but initialState has dynamic behaviour and it felt like incredibly dangerous and dev-unpredictable code.

Instead the true initial state now gets cached for the purpose that the original code indicated it was for.

thisandagain:
Confirmed fixed on Nexus 5, Nexus 4, and Galaxy S

www_src/components/basic-element/basic-element.jsx
k88hudson: Would have to check the source code, but using `getInitialState` to reset element state is a common pattern/has no downsides AFAIK. This is fine, although it does open a chance of potential bugs happening from inadvertently modifying the reference to `this.buttonStyle`

www_src/components/basic-element/basic-element.jsx
```diff
@@ -78,31 +85,46 @@ var El = React.createClass({
...
-    var elWrapper = this.getDOMNode();
-
-    // FIXME: TODO: React has references, so using querySelector is an immediate red flag
-    var elStyleWrapper = elWrapper.querySelector('.style-wrapper');
-    var elButton = elWrapper.querySelector('.meta-button');
-
-    var boundingBox = elStyleWrapper.getBoundingClientRect();
-    var buttonStyle = this.getInitialState();
-
+    var boundingBox = this.refs.styleWrapper.getDOMNode().getBoundingClientRect();
```
k88hudson: :+1:

www_src/components/basic-element/basic-element.jsx
```diff
@@ -115,113 +137,73 @@ var El = React.createClass({
...
-    var state = this.getInitialState();
-
-    var className = classes('el', 'el-' + this.props.type, {
-      touchactive: this.state.touchactive,
-      current: this.props.isCurrent
-    });
+  generateDestinationButton: function() {
+    return (
+      <div className="el-container" key={this.props.key + '-2'}>
+        {/* using onTouchEnd because onClick doesn't work for some reason...touchhandler.js preventing it? */}
```
k88hudson: Might be if touchhandler.js prevents default on any events, that would cancel onClick

www_src/pages/page/flattening.js
```diff
@@ -0,0 +1,16 @@
+var types = require('../../components/basic-element/basic-element.jsx').types;
```
k88hudson: IMO these should be methods on `Spec`, not as a mixin.

k88hudson: that way they can be tested and modified along with their corresponding instance methods

www_src/components/basic-element/basic-element.jsx
Pomax: Agreed: if it's to reset the component, that's a perfect use for getInitialState. That's not what the code did at all, though, it pretended getInitialState was a style definition

k88hudson:
@Pomax Awesome work! Just a few small comments

I am noticing unfortunately that it's very difficult to get a "tap"... is there any way to increase the allowed distance?

www_src/components/basic-element/basic-element.jsx
k88hudson: Ah, got it! thanks

www_src/components/basic-element/basic-element.jsx
Pomax: this isn't actually mine, I think @gvn added this at some point, when he kept running into his clicks getting intercepted somewhere

www_src/components/basic-element/basic-element.jsx
k88hudson: Thanks for working through this :+1: Refactor looks great

www_src/components/basic-element/basic-element.jsx
k88hudson: ah ok, no worries then!

www_src/pages/page/flattening.js
Pomax: also fine: would you be okay with a followup for that?

k88hudson: sure thing :)

Pomax:
hm, I'm able to tap fairly consistently on my device, there's no distance threshold at the moment, but that would be a good idea to look at as immediate follow up. Does that sound reasonable?

Pomax:
https://github.com/mozilla/webmaker-android/issues/2151 files for the flatten/expand refactor followup

Pomax:
https://github.com/mozilla/webmaker-android/issues/2152 filed for the tap-with-some-margin-for-error followup

k88hudson:
Thanks for filing the follow-ups.

This looks like an R+ to me :fireworks: !

thisandagain:
Testing in progress ... :iphone: :iphone: :iphone: :iphone: :iphone:

thisandagain:
This is great. R+ :panda_face:

Issue #2094, opened 2015-06-05
Add tabbing to login form


k88hudson:
I'm not sure why pressing Next/Enter on the android keyboard doesn't work.

errietta:
I think android doesn't like giving next buttons in web forms, I tried a simple form & the browser didn't show a next button. See also http://stackoverflow.com/a/6632329

A 'required' attribute may help focus the next required input though..

PR #2137, opened 2015-06-10
Simulate "tab" when pressing Android "done" button

thisandagain:
This was really difficult to nail down but I think the log-in and sign-up forms are feeling more solid / usable now. I implemented this in a way that should be re-usable / flexible for us going forward.

Resolves GH-2094

### DIFF
--- a/www_src/lib/keyboard.js
+++ b/www_src/lib/keyboard.js
@@ -1,5 +1,5 @@
 module.exports = {
-  simulateTabPress: function () {
+  focusNextInputByTabIndex: function () {
     var current = document.activeElement;
     var elements = document.getElementsByTagName('input');
     var idx = parseInt(current.getAttribute('tabIndex'), 10) + 1;
--- a/www_src/lib/keyboard.js
+++ b/www_src/lib/keyboard.js
@@ -6,7 +6,9 @@ module.exports = {
 
     for (var i = 0; i < elements.length; i++) {
       var t = parseInt(elements[i].getAttribute('tabIndex'), 10);
-      if (t === idx) elements[i].focus();
+      if (t === idx) {
+        elements[i].focus();
+      }
     }
   }
 };
--- a/www_src/lib/keyboard.js
+++ b/www_src/lib/keyboard.js
@@ -1,0 +1,12 @@
+module.exports = {
+  simulateTabPress: function () {
+    var current = document.activeElement;
+    var elements = document.getElementsByTagName('input');
+    var idx = parseInt(current.getAttribute('tabIndex'), 10) + 1;
+
+    for (var i = 0; i < elements.length; i++) {
+      var t = parseInt(elements[i].getAttribute('tabIndex'), 10);
+      if (t === idx) elements[i].focus();
+    }
+  }
+};
--- a/www_src/pages/login/form-input.jsx
+++ b/www_src/pages/login/form-input.jsx
@@ -3,13 +3,21 @@ var React = require('react');
 var FormInput = React.createClass({
   getDefaultProps: function () {
     return {
-      type: 'text'
+      type: 'text',
+      tabIndex: 0
     };
+  },
+  checkForReturn: function (e) {
+    if (e.keyCode === 13) {
+      if (typeof this.props.onReturn === 'function') {
+        this.props.onReturn();
+      }
+    }
   },
   render: function () {
     return (<div className="form-group">
       <label for={this.props.name}>{this.props.label}</label>
-      <input name={this.props.name} type={this.props.type} valueLink={this.props.valueLink} required={this.props.required} />
+      <input name={this.props.name} type={this.props.type} tabIndex={this.props.tabIndex} onKeyDown={this.checkForReturn} valueLink={this.props.valueLink} required={this.props.required} />
       <div className="error" hidden={!this.props.errors}>
         {this.props.errors && this.props.errors.join(' ')}
       </div>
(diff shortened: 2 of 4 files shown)


www_src/lib/keyboard.js
```diff
@@ -0,0 +1,14 @@
...
+    var current = document.activeElement;
+    var elements = document.getElementsByTagName('input');
+    var idx = parseInt(current.getAttribute('tabIndex'), 10) + 1;
+
+    for (var i = 0; i < elements.length; i++) {
+      var t = parseInt(elements[i].getAttribute('tabIndex'), 10);
+      if (t === idx) {
+        elements[i].focus();
+      }
+    }
```
Pomax: sneaky trick:
```
elements.forEach((e,t) => {
  t = e.getAttribute('tabIndex') | 0;
  if (t === idx) {
    e.focus();
  }
});
```

www_src/pages/login/form-input.jsx
```diff
@@ -3,13 +3,21 @@ var React = require('react');
...
   getDefaultProps: function () {
     return {
-      type: 'text'
+      type: 'text',
+      tabIndex: 0
     };
   },
+  checkForReturn: function (e) {
+    if (e.keyCode === 13) {
+      if (typeof this.props.onReturn === 'function') {
```
Pomax: can this even not be a function? If so we can probably just do `if (this.props.onReturn) { this.props.onReturn(); }`

www_src/lib/keyboard.js
Pomax: even sneakier trick:
```
// this is a nodelist *live view*, and always kept up to date
var tabElements = document.querySelectorAll("*[tabindex]");

function next(from) {
  var arr = Array.slice.call(tabElements);
  var curpos = arr.indexOf(from);
  var nextpos = curpos + 1;
  if  (nextpos >= arr.length) { nextpos = 0; }
  return arr[nextpos >= arr.length ? 0 : nextpos];
}
```

www_src/pages/login/form-input.jsx
thisandagain: Because it is a prop set by the parent, it could be `undefined` or just about anything another dev specifies.

www_src/lib/keyboard.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  simulateTabPress: function () {
+    var current = document.activeElement;
+    var elements = document.getElementsByTagName('input');
```
Pomax: technically, not enough. `<select>` and `<textarea>` and friends also allow tabindex for cycling through. `document.querySelectorAll('input, select, textarea, button, a, iframe');` if we want to be *really* complete, but leaving off "a" and "iframe" probably makes sense.

www_src/lib/keyboard.js
```diff
@@ -0,0 +1,14 @@
...
+    var idx = parseInt(current.getAttribute('tabIndex'), 10) + 1;
```
Pomax: vanilla JS trick: `var idx = current.getAttribute('tabIndex') | 0 + 1`, which force-coerces to a 32 bit decimal int. Mind you, jslint *will* complain about that by default (not sure if we care =)

www_src/lib/keyboard.js
```diff
@@ -0,0 +1,14 @@
...
+
+    for (var i = 0; i < elements.length; i++) {
+      var t = parseInt(elements[i].getAttribute('tabIndex'), 10);
+      if (t === idx) {
```
Pomax: technical nit: this isn't true. In HTML, DOM ordering matters too, so you can have 100 elements with the same tab index, and the "next" element to pick in absence of a higher tabindex is simply the next in line. That's way more than we probably need, but it's worth noting in the code that this is purely for testing, and an incomplete implementation

www_src/pages/login/sign-up.jsx
```diff
@@ -47,6 +48,8 @@ var SignUp = React.createClass({
     {
       name: 'username',
       label: 'Username',
+      type: 'email',
```
Pomax: should this be "email", or do we have a username specific type?

www_src/pages/login/form-input.jsx
Pomax: `undefined` and `false` are fine. If that's all they can be, no need to be overly safe. If we do explicitly the property, with "not a function", then this code is better.

www_src/lib/keyboard.js
thisandagain: In a mobile context all we really care about is `input` and `textarea`. Will add `textarea` for completeness.

www_src/lib/keyboard.js
thisandagain: Why would I use a coercion here when it is harder to read + throws a lint error?

www_src/pages/login/form-input.jsx
thisandagain: Not sure I understand. From an ergonomics perspective, a dev can set `onReturn` to any value or type. Because hitting the return button while testing is something that could be easily overlooked, checking to ensure that `onReturn` is a function seems perfectly reasonable.

www_src/pages/login/sign-up.jsx
thisandagain: Trick to avoid keyboard suggestions on Android (which cause a noticeable "glitch" when switching between fields) and not wanted when typing a username.

www_src/lib/keyboard.js
thisandagain: https://github.com/airbnb/javascript#type-casting--coercion

www_src/pages/login/sign-up.jsx
k88hudson: Can you add this as a comment?

k88hudson:
R+

Issue #2134, opened 2015-06-10
Unable to access tinker mode for border colors


thisandagain:
From @LauraReynal 
```bash
[6/9/15, 5:52:17 PM] Laura de Reynal: there is a bizarre thing with the borders though
[6/9/15, 5:52:19 PM] Laura de Reynal: its still here
[6/9/15, 5:53:33 PM] Laura de Reynal: if you want to reach tinker mode for borders
[6/9/15, 5:53:42 PM] Laura de Reynal: then you have to hit a normal color first
[6/9/15, 5:54:09 PM] Laura de Reynal: and results are not guaranteed
```

STR:
- Create image element
- Attempt to press "tinker mode" button for border

Expected:
- User enters "Tinker Mode"

Actual:
- Tinker mode button is disabled until you select another color

k88hudson:
Resolved by commit https://github.com/k88hudson/webmaker-android/commit/3d4bd57712328b4281e40b3c70b4cb2e3782dcf0

LauraReynal:
Not working with this morning APK> can you guys try?

LauraReynal:
@thisandagain @k88hudson

k88hudson:
Hmm, seems to be OK for me. @LauraReynal could you maybe re-download the latest nightly just in case?

xmatthewx:
Ok on my end.

k88hudson:
@secretrobotron confirms it's working for him

PR #2139, opened 2015-06-10
[#2134] Launch tinker mode even if no edits were made yet

k88hudson:


k88hudson:
@thisandagain R?

thisandagain:
Only issue I'm seeing is that the "Transparency" slider is set to `1` when you enter Tinker Mode (I'm testing border color on an image) even though the value shown is `0`. From a user experience perspective, I'm wondering if we should keep the default as one and update the state. Without that the color selector feels broken even though the user intent is most likely to pick a color. Make sense? Thoughts?

k88hudson:
@thisandagain good catch! this was actually a bug in our `Range` component

thisandagain:
W00t! Thanks for the quick fix. R+

PR #2140, opened 2015-06-10
Add edit to play

### LINKED ISSUE #2128: Add "edit" to play activity
Add "edit" to a project's more action menu in the play view. Edit option should appear for authors only.

errietta:
Closes #2128

### DIFF
--- a/app/src/main/java/org/mozilla/webmaker/activity/Play.java
+++ b/app/src/main/java/org/mozilla/webmaker/activity/Play.java
@@ -1,13 +1,17 @@
 package org.mozilla.webmaker.activity;
 
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.net.Uri;
 import android.os.Bundle;
 import android.view.Menu;
 import android.view.MenuItem;
 import org.json.JSONException;
+import org.json.JSONObject;
 import org.mozilla.webmaker.R;
 import org.mozilla.webmaker.WebmakerActivity;
+import org.mozilla.webmaker.javascript.WebAppInterface;
+import org.mozilla.webmaker.router.Router;
 import org.mozilla.webmaker.util.Share;
 
 public class Play extends WebmakerActivity {
@@ -15,9 +19,27 @@ public class Play extends WebmakerActivity {
         super("project", R.id.page_layout, R.layout.page_layout, R.menu.menu_play);
     }
 
+    protected String sessionId;
+    protected String projectAuthor;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
+        /**
+         * Get the id of the logged in user from the sharedPreferences as well as the id of the
+         * maker of the app. These strings are later compared to give an 'edit' button to the
+         * app creator.
+         **/
+        try {
+            SharedPreferences userSession = getSharedPreferences(WebAppInterface.USER_SESSION_KEY, 0);
+            JSONObject session = new JSONObject(userSession.getString("session", ""));
+
+            sessionId = session.getJSONObject("user").get("id").toString();
+            projectAuthor = routeParams.get("user").toString();
+        } catch (JSONException e) {
+            // Do nothing.
+        }
 
         // Set "mode" in route params
         try {
@@ -32,15 +54,29 @@ public class Play extends WebmakerActivity {
 
         // Set route params if valid deep link
         if (data == null) return;
-        final String user = data.getQueryParameter("user");
+        projectAuthor = data.getQueryParameter("user");
         final String project = data.getQueryParameter("project");
-        if (user == null || project == null) return;
+
+        if (projectAuthor == null || project == null) return;
         try {
-            routeParams.put("user", user);
+            routeParams.put("user", projectAuthor);
             routeParams.put("project", project);
         } catch (JSONException e) {
             // do nothing
         }
+    }
+
+    @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        MenuItem editItem = menu.findItem(R.id.action_edit);
+
+        if (editItem == null || sessionId == null) return true;
+
+        if (projectAuthor.equals(sessionId)) {
+            editItem.setVisible(true);
+        }
+
+        return true;
     }
 
     @Override
@@ -68,6 +104,9 @@ public class Play extends WebmakerActivity {
                 return true;
             case R.id.action_remix:
                 view.load("javascript: window.createRemix && window.createRemix()", null);
+                return true;
+            case R.id.action_edit:
+                Router.sharedRouter().open("/users/" + userId + "/projects/" + id);
                 return true;
             default:
                 return super.onOptionsItemSelected(item);
--- a/app/src/main/res/menu/menu_play.xml
+++ b/app/src/main/res/menu/menu_play.xml
@@ -10,4 +10,9 @@
           android:title="@string/action_remix"
           android:orderInCategory="100"
           app:showAsAction="always" />
+    <item android:id="@+id/action_edit"
+          android:title="Edit"
+          android:orderInCategory="100"
+          app:showAsAction="always"
+          android:visible="false" />
 </menu>


app/src/main/java/org/mozilla/webmaker/WebmakerActivity.java
```diff
@@ -32,6 +35,15 @@ public WebmakerActivity(String pageName, int id, int layoutResID, int menuResId)
     protected void onCreate(Bundle savedInstanceState) {
         routeParams = new JSONObject();
 
+        SharedPreferences userSession = getSharedPreferences(WebAppInterface.USER_SESSION_KEY, 0);
```
thisandagain: This is going to cause a fairly sizable performance hit `onCreate` of each Activity. I'd rather isolate this to just the `Play` activity.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -70,6 +108,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
             case R.id.action_remix:
                 view.load("javascript: window.createRemix && window.createRemix()", null);
                 return true;
+            case R.id.action_edit:
```
thisandagain: :+1:

app/src/main/res/menu/menu_play.xml
```diff
@@ -10,4 +10,10 @@
           android:title="@string/action_remix"
           android:orderInCategory="100"
           app:showAsAction="never" />
+    <item android:id="@+id/action_edit"
```
thisandagain: :+1:

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +46,42 @@ protected void onCreate(Bundle savedInstanceState) {
     }
 
     @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        if (routeParams == null) {
```
thisandagain: We can remove this as per note above.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +46,42 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            return true;
+        }
+
+        MenuItem item = menu.findItem(R.id.action_edit);
```
thisandagain: You should use a more descriptive name than `item`. Also, I'd recommend you add a line below this to also include the call to `SharedPreferences`. Then, you can do:
```java
if (editMenuItem == null || sessionId == null) return true;
```

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +46,42 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            return true;
+        }
+
+        String mode = "";
+        JSONObject session;
+        String creator = "";
+        String user = "";
+
+        try {
+            mode = routeParams.get("mode").toString();
```
thisandagain: You know that mode will always be "play" because you are in the `Play` activity.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +46,42 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            session = new JSONObject(routeParams.getString("session"));
+
+            user = session.getJSONObject("user").get("id").toString();
```
thisandagain: Instead of getting the user id and project id from routeParams again here, I'd recommend you move the getters from the `onOptionsItemSelected` block up to `onCreate` and make them available to the entire class as `protected` (final) vars.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +46,42 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            creator = routeParams.getString("user");
+        } catch (JSONException e) {
+            e.printStackTrace();
+        }
+
+        if (mode.equals("play") && creator.equals(user) && !creator.equals("")) {
+            item.setVisible(true);
+        } else if (mode.equals("play")) {
+            item.setVisible(false);
```
thisandagain: You don't need to explicitly `setVisible(false)` as it is invisible by default as per your settings in `menu_play.xml`

thisandagain:
Totally awesome. This is working on-device for me. Some performance and notes on the implementation but this is a great start.

app/src/main/res/menu/menu_play.xml
```diff
@@ -10,4 +10,10 @@
...
+          android:title="Edit"
+          android:orderInCategory="100"
+          app:showAsAction="never"
```
ryanw-se: This will need to be changed from ```never``` to ```always```, to reflect a bug that @thisandagain found last night involving the ellipse menu not showing.

k88hudson:
@errietta mind if we continue this after the beta launch?

errietta:
@k88hudson  aww :cry:  when is that?

k88hudson:
@errietta just don't want to land new features until after Friday

errietta:
@thisandagain Re-R when you can? :fire:

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -15,10 +19,23 @@ public Play() {
         super("project", R.id.page_layout, R.layout.page_layout, R.menu.menu_play);
     }
 
+    protected String login_username;
```
thisandagain: Please use camelCase for variable names.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -15,10 +19,23 @@ public Play() {
...
+    protected String app_creator;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+        try {
```
thisandagain: This block needs a comment to explain what is happening.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -15,10 +19,23 @@ public Play() {
...
+            SharedPreferences userSession = getSharedPreferences(WebAppInterface.USER_SESSION_KEY, 0);
+            JSONObject session = new JSONObject(userSession.getString("session", ""));
+
+            login_username = session.getJSONObject("user").get("id").toString();
+            app_creator = routeParams.get("user").toString();
+        } catch (JSONException e) {
+            e.printStackTrace();
```
thisandagain: You can `// do nothing` here. No need to print the stack trace is a user session does not exist (which will be common).

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -44,6 +62,20 @@ protected void onCreate(Bundle savedInstanceState) {
     }
 
     @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+        MenuItem edit_item = menu.findItem(R.id.action_edit);
+
+        if (edit_item == null || login_username == null)
```
thisandagain: Nitpick: Use `{}` consistently. The style below is what we generally use throughout the app.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -15,10 +19,23 @@ public Play() {
         super("project", R.id.page_layout, R.layout.page_layout, R.menu.menu_play);
     }
 
+    protected String login_username;
+    protected String app_creator;
```
thisandagain: Probably best to be consistent about how we reference the "app_creator" and project which would be: `projectAuthor` and `projectId`.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
thisandagain: You are storing the user `id` not the `username`. Something like `sessionId` might make more sense here.

thisandagain:
Nice! This is working well and looking super close. Some nitpicks and then I think this is ready for final testing.

errietta:
@thisandagain R?again!

ryanw-se:
Woot :dog:

thisandagain:
R+ :+1:

PR #2146, opened 2015-06-10
[#1966] Add a few missing loading states

k88hudson:
Mostly complete, just a couple of things missed


k88hudson:
@thisandagain R?

Pomax:
This is likely affected by https://github.com/mozilla/webmaker-android/pull/2136, if possible let's land that first, then rebase this, rather than the other way around (it would be a ton of work to fix 2136 back up)

thisandagain:
R+

Issue #1833, opened 2015-05-12
[MIGRATED] Implement multi-line text editor


flukeout:
### Implementation

Replace single-line text editor with multi-line text editor

**STATUS**: Design in progress

#### Original notes (filed by flukeout):

Steps to reproduce
* Add a text element, and edit it
* Change the text to "asdf asdf asdf asdf asdf asdf asdf asdf asdf asdf" or "asdfasdf asdfasdf asdfasdf asdfadsf asdfadsf" or maybe even "asdfjjak kdjf ldkjf k dlfkj dlkfj adsklf jas"
* Or anything that has a lot of words and spaces!
* You'll notice that the preview will wrap the words across multiple lines
* Confirm the changes and return to the Page Editor
* The Text Element displays the text you typed all on one line

I know we're replacing the text editor, but this issue may still persist so we should be aware of it. We need to figure out how to handle the UX of wrapping and line breaks.

Pomax:
should be fixed with https://github.com/mozilla/webmaker-android/pull/1877

secretrobotron:
Just tried this with latest nightly.
Text doesn't wrap in editor either, and is always slightly misaligned.

secretrobotron:
Blocks #1665

secretrobotron:
cc @k88hudson

k88hudson:
@vazquez can you add a comp/design spec for the multiline text editor?

Notes from our discussion:

* text editor should be vertically centered in preview window
* pressing enter should add a line break instead of exiting the editor
* users should use the hide keyboard button to dismiss keyboard

Note that if the editor is styled appropriately, there shouldn't really be a reason to ever hide it/replace it with a non-editable version (it can just always be shown)

k88hudson:
cc @Pomax

xmatthewx:
Will we implement resize of text container in the page view? Getting that to feel right is the bit that concerns me.

k88hudson:
@xmatthewx If you think it would bee easier, having the textarea fill the entire preview and have text top-aligned would be better than nothing -- however, I was hoping we could go for something like this http://codepen.io/desandro/pen/gICqd

vazquez:
So something like this:

![text options multiline](https://cloud.githubusercontent.com/assets/1119821/8064474/f733e790-0ea8-11e5-8597-a4d1782a7909.png)

k88hudson:
@vazquez awesome, thanks!

xmatthewx:
[...]

PR #2149, opened 2015-06-10
Interim solution for multiline text [#1833]

k88hudson:
I talked through this with @secretrobotron, and we came up with an interim solution. This is not ideal solution, but slightly improves the UX and is a minimal code change

### DIFF
--- a/www_src/components/basic-element/basic-element.less
+++ b/www_src/components/basic-element/basic-element.less
@@ -60,6 +60,7 @@
 
   // Text
   &.el-text {
+    max-width: 320px;
     p {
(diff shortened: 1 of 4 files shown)


gvn:
So I think this is good for now. Maybe I'm just missing something, but you can't create a line break manually, right? It just wraps the text after it gets beyond a certain length?

k88hudson:
@gvn that's correct, allowing for manual line breaks is more work than we're able to complete atm; @Pomax is going to implement a follow-up with the actual user experience

gvn:
Cool. R+!

Issue #2178, opened 2015-06-12
Tapping card in style guide crashes Webmaker


cadecairos:
presumably because it's a fake project?

```
06-12 09:14:51.476  19647-19647/org.mozilla.webmaker E/AndroidRuntime﹕ FATAL EXCEPTION: main
    Process: org.mozilla.webmaker, PID: 19647
    org.mozilla.webmaker.router.exception.RouteNotFoundException: No route found for url /project/123
            at org.mozilla.webmaker.router.Router.paramsForUrl(Router.java:276)
            at org.mozilla.webmaker.router.Router.open(Router.java:189)
            at org.mozilla.webmaker.router.Router.open(Router.java:177)
            at org.mozilla.webmaker.javascript.WebAppInterface$2.run(WebAppInterface.java:206)
            at android.os.Handler.handleCallback(Handler.java:739)
            at android.os.Handler.dispatchMessage(Handler.java:95)
            at android.os.Looper.loop(Looper.java:155)
            at android.app.ActivityThread.main(ActivityThread.java:5696)
            at java.lang.reflect.Method.invoke(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:372)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1028)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:823)
```

ryanw-se:
@cadecairos I believe this is because when the login system landed, the routes were changed to reflect a new param ```/user``` and this one doesn't appear to have been updated. Good catch!

thisandagain:
Thanks @ryanw-se !

PR #2180, opened 2015-06-12
[#2178] Fix project card in StyleGuide

ryanw-se:
Fixes issue #2178 which was causing the app to crash when you tapped on the project card in style guide, due to a invalid route that wasn't updated for the new user param and a slight typo that existed with the current param.


cadecairos:
R+, fixes the crash on my device.

Issue #2174, opened 2015-06-12
Set destination green button does not do anything 


LauraReynal:
- add button

- click on set link destination 

You are then brought to the zoom out page to choose your destination 

- click on your destination 
at this point, you either click on the green button below or you want to click on the back arrow to save. 
nothing seems to be happening when you click on the green button that says "set link destination"

LauraReynal:
4 users right now are confirming that links are not working

Pomax:
PR merged.

PR #2186, opened 2015-06-12
fixing link destination functionality

gvn:
@Pomax looks like some of your refactoring on link editing caused a regression... :sob:


gvn:
Closes #2174

xmatthewx:
Does this also address that set destination button stopped working in the page view? 

Also, in destination editor, previous destinations are set on new buttons (probably not a critical bug if everything else still works).

gvn:
Yes, set destination should work from both locations now.

Do you want to file a bug for the other thing?

Issue #2066, opened 2015-06-02
Auto refresh Make view on project exit


xmatthewx:
When I create a new project, I expect to see that project in Make when I exit the project. Currently, I need to manually refresh.

xmatthewx:
This also impacts remixes as noted by @gvn in https://github.com/mozilla/webmaker-android/pull/2141#issuecomment-110863555. A user will likely assume that their remix failed. 

Refresh would also be helpful for seeing updates to thumbnails and titles. Titles in particular feel like they failed to save. 

Seems important for v1.

gvn:
Yeah, basically I think Make needs to always pull fresh data when it is focused. I'm not entirely sure how to get that to happen since it isn't being destroyed and rebuilt. Do we have some kind of focus event from Android that we can tap into? @k88hudson @thisandagain

thisandagain:
Yup. We get Activity lifecycle events from the `MainActivity`. I can take this one.

gvn:
:+1:

PR #2193, opened 2015-06-12
[#2066] Refresh make view with updates

k88hudson:
This refreshes after pause, resume, and changes in which tab is active -- do you think that's too often?


k88hudson:
@thisandagain R? [[flagged!]](http://plzkthx.herokuapp.com)

thisandagain:
On tab selected is too frequently (particularly when we have pull-to-refresh). Remove that and I think this is good to go.

k88hudson:
@thisandagain is this ok?

thisandagain:
Working well! R+

PR #2197, opened 2015-06-12
Fixes the weird Android text duplication. For now

Pomax:
fixes https://github.com/mozilla/webmaker-android/issues/2050, although as a hack. I suspect Android will keep doing this forever, but if it *doesn't*, we will need to revisit this hackfix.


thisandagain:
Nice. This is working well.

Issue #1922, opened 2015-05-20
[MIGRATED] Implement Error states


thisandagain:
We need to have error states when creating / editing / deleting on projects, pages, and elements.

flukeout:
**List of possible errors**

* [ ] Username too short or too long
* [ ] Username contains illegal characters
* [ ] Failed to create a new project
* [ ] Failed to delete a project
* [ ] Failed to load a project
* [ ] Failed to load list of projects in Make or Discover views
* [ ] Failed to load an image in a project
* [ ] Failed to load a page within a project
* [ ] Failed to upload an image when editing a project

k88hudson:
Here is a list for each view:

### General

* [ ] User loses/has no network connection
* [ ] Network request times out

### Discover

* [ ] Discover gallery fails to load
* [ ] Discover gallery has no items

### Make

* [ ] Make gallery fails to load
* [ ] Make gallery has no items
* [ ] Create new project fails

### Project

* [ ] Project fails to load/not found
* [ ] Project is empty, attempt to create first page fails*
* [ ] Create new page fails
* [ ] Delete page fails

*Note: This is when a project is first created, has no pages, and the first page is created. This may need some specific behaviour here, like a prompt to re-try creating the first page... unless we can simplify creating the first page to a single request https://github.com/mozilla/api.webmaker.org/issues/53

### Page

* [ ] Page fails to load/not found
* [ ] Position update fails (scale, translation, etc.)
* [ ] Create new element fails
* [ ] Delete element fails

### Element editor

* [ ] Element fails to load/not found
* [ ] Style/attribute update fails

### Tinker mode

* [ ] Element fails to load/not found
* [ ] Style/attribute update fails

k88hudson:
SO MUCH ERROR :warning:

flukeout:
@k88hudson I started some work in [The Error States Diagram](https://dl.dropboxusercontent.com/u/109902/tiles/error-states.pdf) - please have a look over it and let me know your thoughts.

flukeout:
@vazquez Has already done some work on [No internet in Discover mode](https://www.dropbox.com/s/m3mdb8782smp6iz/Screenshot%202015-05-21%2015.44.42.png?dl=0) state.

xmatthewx:
@flukeout - your spec on these states looks great. You've nicely divided things we should report via
 - the view background
[...]

PR #2215, opened 2015-06-13
Errors

Pomax:
addresses #1922

### DIFF
--- a/www_src/pages/make/make.jsx
+++ b/www_src/pages/make/make.jsx
@@ -1,10 +1,12 @@
 var React = require('react');
-var render = require('../../lib/render.jsx');
-var api = require('../../lib/api.js');
-var Card = require('../../components/card/card.jsx');
-var Loading = require('../../components/loading/loading.jsx');
+var api = require('../../lib/api');
 var router = require('../../lib/router');
 var dispatcher = require('../../lib/dispatcher');
+var reportError = require('../../lib/errors');
+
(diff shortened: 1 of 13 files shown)


thisandagain:
In most instances I'm having trouble triggering an error state. @Pomax how were you testing this?

www_src/pages/make/make.jsx
```diff
@@ -24,15 +26,14 @@ var Make = React.createClass({
     this.load();
   },
   onError: function (err) {
-    console.error(err);
+    reportError("Error loading project", err);
```
thisandagain: This should be "Error loading projects" (plural)

www_src/pages/make/make.jsx
```diff
@@ -24,15 +26,14 @@ var Make = React.createClass({
...
     this.setState({loading: false});
   },
   onEmpty: function () {
-    console.log('No projects found');
+    reportError('No projects found');
```
thisandagain: We should not show an error when no projects exist as it makes for a poor first-time-use.

thisandagain:
This is great. I love how clean the `snackbar` component turned out. Two really minor comments, otherwise R+

www_src/pages/make/make.jsx
Pomax: :+1:

thisandagain:
RRRRRRR+

PR #2231, opened 2015-06-15
Fix display of transparency on tinker and image editors

k88hudson:
Represent opacity as a percentage on normal editors, tinker editor renames transparency to opacity.


gvn:
R+!

PR #2236, opened 2015-06-15
Use variables for zindex, make action buttons go above meta button

k88hudson:
No bug, but I noticed that it's impossible to add a new element if a link element is currently selected on the page editor

### DIFF
--- a/www_src/pages/element/element.less
+++ b/www_src/pages/element/element.less
@@ -2,65 +2,6 @@
   color: @slate;
   height: 100%;
   overflow-y: scroll;
-
-  .overlay {
-    position: fixed;
-    top: 0;
-    left: 0;
-    height: 100%;
-    width: 100%;
-    z-index: 998;
-    background: #1A2237;
-    opacity: 0;
-    visibility: hidden;
-    transition: all 0.2s ease-in-out;
-    &.active {
-      visibility: visible;
-      opacity: 0.7;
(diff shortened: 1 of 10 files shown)


www_src/pages/element/element.less
```diff
@@ -2,65 +2,6 @@
   color: @slate;
   height: 100%;
   overflow-y: scroll;
-
```
k88hudson: unused duplicated code

www_src/variables.less
```diff
@@ -0,0 +1,10 @@
+// Action/+ menu
+@ZINDEX_ACTION_OVERLAY: 999;
+
+// Set link destination button
+@ZINDEX_META_BUTTON: @ZINDEX_ACTION_OVERLAY - 2;
+
+// Loading indicator
+@ZINDEX_LOADING: 10000;
+
+@ZINDEX_MAX: 2147483647; /* signed 32bit maxint */
```
gvn: Should really be `@ZINDEX_POMAX` :stuck_out_tongue_winking_eye:

gvn:
Seems good from my general smoke testing. Since this is a broad change and we have no integration tests I'd suggest getting a 2nd reviewer! :+1:

k88hudson:
@gvn thanks!

@Pomax mind giving this a quick check-over?

PR #2237, opened 2015-06-15
bigger default text (closes #2181)

### LINKED ISSUE #2181: Default text is too small to scale up
When users add new text, it's sometimes too small to see and pinch to scale up. Let's set new defaults: Increase font-size 50% and set font-weight to bold.

gvn:


thisandagain:
R+ :smile:

PR #2239, opened 2015-06-15
Added react-intl and testing trivial example

k88hudson:
I figured I would put up the JS strategy in progress for y'all to comment if you have ideas.

So far, `react-intl` seems like a pretty good solution to me; I'm planning on passing in the locale and strings from Android via the bridge.

### DIFF
--- a/www_src/lib/render.jsx
+++ b/www_src/lib/render.jsx
@@ -4,7 +4,18 @@ var ModalConfirm = require('../components/modal-confirm/modal-confirm.jsx');
 var ModalSwitch = require('../components/modal-switch/modal-switch.jsx');
 var Snackbar = require('../components/snackbar/snackbar.jsx');
 
+var intlData = {
+  locales: ['en-US'],
+  messages: {
+    sign_in: 'Sign In',
+    sign_in_error: 'Looks like there might be a problem with your username or password.',
+    dont_have_account: 'Don\'t have an account?',
+    join_webmaker: 'Join Webmaker'
+  }
+};
+
 var Base = React.createClass({
+  mixins: [require('react-intl').IntlMixin],
   onResume: function () {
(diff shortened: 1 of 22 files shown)


www_src/lib/render.jsx
```diff
@@ -3,7 +3,18 @@ var Spindicator = require('../components/spindicator/spindicator.jsx');
 var ModalConfirm = require('../components/modal-confirm/modal-confirm.jsx');
 var ModalSwitch = require('../components/modal-switch/modal-switch.jsx');
 
+var intlData = {
```
k88hudson: this will be from android

k88hudson:
cc @thisandagain / @pomax / @gvn

gvn:
Seems like a good approach. When running on a desktop browser will localized strings just be empty?

alicoding:
So, I'm trying to understand what would be the benefit for us using XML for our string resource here? Is it just so we can use it in both React and on Android side? I really like the idea of us using ICU format for translation since it makes it really easy to maintain and readable. Also, it is supported by default with React-Intl.

What do you think about separating them totally in this case? For example:

Continue using XML in Webmaker-android and take of all the webmaker-core strings. Use JSON for webmaker-core - I like the idea of sharing, but having to build the file and share one giant resource file is hard to maintain.

Thought? @Pomax @thisandagain @k88hudson @gvn

toolness:
<a href="https://github.com/mozilla/webmaker-core/issues/193"></a>
This issue has been moved to https://github.com/mozilla/webmaker-core/issues/193.
Please don't comment on it here.

toolness:
Sorry, my migration script for https://github.com/mozilla/webmaker-core/issues/23 accidentally migrated this, even though it wasn't supposed to. Just undid it!

PR #2244, opened 2015-06-15
ignoring link button actions in Link mode (resolves #2238)

### LINKED ISSUE #2238: Disable links in 'set destination' view
Links placed by authors function on the set destination view. Try and select a page as your destination, and you might accidentally trigger the button. The view pans and adds the flag on the unintended destination.

I noticed this while making a table of contents page with a ton of buttons. It's nearly impossible to set this as a destination.

gvn:


thisandagain:
R+

PR #2253, opened 2015-06-16
add webkit transform to support screenshots

k88hudson:
This was a regression in desktop, i'm adding it here so the regression doesn't happen again


gvn:
R+!

PR #2263, opened 2015-06-17
Use react-imageloaders prerender property to prevent 0 height thumbnails

### LINKED ISSUE #2207: [MIGRATED] Discover gallery has 0 height thumbnails
After creating a new account and opening discovery for the first time, I was somehow able to get to a state where all the thumbnails were 0 height. After refreshing, they were fine. Not able to reproduce this a second time.

cadecairos:
Fixes #2207 

This "works" but lets see if we can find a real solution to the strange loading behaviour before going with this.


thisandagain:
Great! R+

Issue #2256, opened 2015-06-16
Big Button Effect


gvn:
Currently you can "lock yourself out" of editing a page by either filling it with a giant button or too dense a group of smaller buttons. This is caused by buttons blocking the action of tapping a page to select it.

We need to come up with a way to reduce or remove this competition between UI.

As a short term solution I'd recommend at least limiting the maximum size of a button...

gvn:
CC: @xmatthewx @vazquez @LauraReynal @thisandagain

xmatthewx:
Works for me. It's easy to imagine wanting a full-width button. How about we allow full width and restrict height to 50% of page?

flukeout:
I ran into this in issue #2260 on the Fire Island stock project. The first page has a ton of buttons, and it's hard to tap it to zoom in without being navigated away.

xmatthewx:
Stock project? I'll take that as a compliment :wink:

gvn:
We landed a bandaid for this, so I'm going to take it out of this sprint but leave it open for revisiting later.

flukeout:
@gvn What's the bandaid?

xmatthewx:
tracking in #2260

PR #2265, opened 2015-06-17
upping min button size; setting max size

gvn:
Related: #2256


www_src/components/basic-element/basic-element.jsx
```diff
@@ -18,7 +18,8 @@ var BasicElement = React.createClass({
       link: require('./types/link.jsx')
     },
     // the minimum size of the long-edge of an element, in pixels.
-    minScaledEdgeLength: 44,
```
gvn: I was able to get things too small to pinchzoom at this width...

thisandagain:
Nice. This is working well. R+

PR #2276, opened 2015-06-17
no highlight for destination pages after clicking a button in Play mo…

### LINKED ISSUE #2261: In Play Mode, navigating to a page via a button causes that page to become "selected"
Steps to reproduce
* Open a project with pages and buttons
* Tap a button
* The view transitions to the destination page
* The destination page now has a blue "selected" border

It shouldn't. There is no selecting pages in Play mode.

gvn:
…de (resolves #2261)


k88hudson:
R+

PR #2277, opened 2015-06-18
Projecttransform rebase

Pomax:
supercedes https://github.com/mozilla/webmaker-android/pull/2264

### DIFF
--- a/www_src/pages/project/transforms.js
+++ b/www_src/pages/project/transforms.js
@@ -1,0 +1,195 @@
+var assign = require('react/lib/Object.assign');
+var calculateSwipe = require('../../lib/swipe.js');
+
+var MAX_ZOOM = 0.8;
+var MIN_ZOOM = 0.18;
+var ZOOM_SENSITIVITY = 300;
+
+/**
+ * Directly manipulate an element's style, rather than changing it through render()
+ */
+function dangerouslySetStyle(element, style) {
+  assign(element.style, style);
+}
+
+/**
+ * utility function for constraining a value
+ */
+function constrain(v,min,max) {
+  return v<min ? min : v>max ? max : v;
+}
+
+/**
+ * Touch handler for projects
+ */
+var handleTouches = function(component) {
+  var node = component.getDOMNode();
+  var master = component.refs.bounding.getDOMNode();
+
+  // administrative values:
+  var didMove = false,
+      startX,
+      startY,
+      startDX = 0,
+      startDY = 0,
+      startDistance,
+      endX,
+      endY,
+      currentX,
+      currentY,
+      currentZoom;
+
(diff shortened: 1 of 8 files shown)


Pomax:
@thisandagain, @xmatthewx can I ask you two to see what you think about the feel of this project view zoom. It's not a strict "to/from center of pinch zooming", but does quite a good job (I think) of letting you find your content by zooming the content "towards and through" the pinch center instead.

xmatthewx:
This PR seems too big and risky to land at the last minute. And, I haven't had a chance to test on device. We should put this on hold unless someone else has an abundance of time and confidence. cc @gvn @k88hudson

Pomax:
emphasis on "seems", the refactor is for the most part a true refactor: the code is exactly the same, just moved to their own files to keep development sane. The only real rewrites are in transform.js

www_src/pages/project/transforms.js
```diff
@@ -0,0 +1,202 @@
...
+    startDY = 0;
+    startDistance = undefined;
+    currentX = undefined;
+    currentY = undefined;
+    currentZoom = undefined;
+  };
+
+  // Swipe handling is supposed to kick in for... something...
+  var handleSwipe = () => {
+    // FIXME: TODO: turned off for now, in part because of Laura's comments, in
```
Pomax: not sure if we want to keep this disabled or reenable it. On a nexus 7, the swiping does pretty much nothing visually different from regular translation

xmatthewx:
@Pomax - good to know. We also "seem" to be short on hands and eyes today, but we can keep the door open and try to land this.

xmatthewx:
That last comment sounds snarky. Not intended. Personally, I will be thrilled if we can include this ... it will greatly improve demos in Whistler.

k88hudson:
@Pomax Really excellent work here. However since this is mostly a refactor/lays the groundwork for future improvements rather than directly impacting the user experience, I would vote to land post-launch or on a separate branch. Does that sound OK to you?

gvn:
I also vote for landing post-launch.

thisandagain:
Thanks for putting this together @Pomax 

I agree with landing this after the Whistler release tomorrow as it introduces quite a bit of risk. Including it in the nightly should give us lots of time to test on-device.

www_src/pages/project/transforms.js
thisandagain: This works in `play` mode currently. Disabling it feels like a regression. Laura's notes were specific to the UK ... users in Bangladesh understood it.

thisandagain:
Just did one more review / walk-through. This is R+ as long as we can re-enable swiping between views in play mode.

thisandagain:
R+ Thanks a ton @Pomax

PR #2282, opened 2015-06-18
bumping to 25 projects shown in discovery tab (closes #2259)

### LINKED ISSUE #2259: More projects in Discover
For Whistler and in general, it would be good to increase the number of projects in Discover. If we can't add pagination to load more just yet, can we bump it up to 25?

cc @secretrobotron

gvn:
Note there aren't 25 featured projects yet...


k88hudson:
R+

PR #2286, opened 2015-06-18
putting back css that was needed after all...

gvn:
Puts back CSS removed in https://github.com/mozilla/webmaker-android/commit/5fb339579dcf152f8301d0bada9d17ead917a3b5 

Removing this code prevented the Change Image functionality from working.


k88hudson:
R+

PR #2318, opened 2015-06-25
Update Gradle & Crosswalk dependencies

ryanw-se:
Should get rid of some lingering bugs and issues that we might have seen previously, I tested this briefly and nothing appears to have broken.


thisandagain:
Working well on Mac and PC test machine / VM. Gradle release notes:
https://docs.gradle.org/1.2/release-notes

PR #2326, opened 2015-06-26
fixes #2325

### LINKED ISSUE #2325: Cannot move element after scaling
Steps to reproduce
* Scale an element, but don't remove the initial finger you placed on the element
* Try to move the element by moving the first finger
* Element doesn't move

it should move. Looks like when the second finger (for scaling) is lifted from the screen, the element forgets that the first finger is still down.

Pomax:
put in a check to make sure we keep in mind whether zero or more finges are on the screen after removing a finger


thisandagain:
Fix confirmed. Great catch @Pomax @flukeout. R+

PR #2372, opened 2015-06-30
Platform Polyfill

thisandagain:
This PR *lightly* introduces a new module called `platform` that attempts to provide a browser polyfill for methods that are generally handled by Android. If the platform provides an interface bound to `window.Platform` it will return it rather than the polyfill. Because this has implications across the codebase, I'm only introducing this polyfill in a few areas for this PR:
- API request caching
- `discover.jsx`
- `make.jsx`
- `render.jsx`

The polyfill covers most of what we use today with a few notable gaps (listed in the lib as `@todo`):
- Camera / image gallery support
- Sharing panel support
- Route params

### DIFF
--- a/www_src/lib/api.js
+++ b/www_src/lib/api.js
@@ -1,10 +1,12 @@
 var xhr = require('xhr');
 var defaults = require('lodash.defaults');
-var {jsonToFormEncoded, parseJSON} = require('./jsonUtils');
-var router = require('./router');
 var assign = require('react/lib/Object.assign');
-var dispatcher = require('./dispatcher');
+
 var config = require('../config');
+var dispatcher = require('./dispatcher');
+var platform = require('./platform');
+var router = require('./router');
+var {jsonToFormEncoded, parseJSON} = require('./jsonUtils');
 
 function api(options, callback) {
 
@@ -41,13 +43,11 @@ function api(options, callback) {
     options.headers.Authorization = 'token ' + router.getUserSession().token;
   }
 
-  // Set cache key
+  // Caching
   var key = 'cache::' + options.method + '::' + options.uri;
-
-  // Use device cache if window.Android is available & options.useCache is true
-  if (window.Android && options.useCache === true && options.method === 'GET') {
+  if (options.useCache === true && options.method === 'GET') {
     console.log('Fetching from cache "' + key + '"');
-    var hit = window.Android.getMemStorage(key, true);
+    var hit = platform.getMemStorage(key, true);
     if (typeof hit === 'string') {
       return callback(null, JSON.parse(hit));
     }
@@ -68,10 +68,8 @@ function api(options, callback) {
       return callback(err);
     }
 
-    // Set cache if window.Android is available
-    if (window.Android) {
-      window.Android.setMemStorage(key, JSON.stringify(body), true);
-    }
+    // Set cache
+    platform.setMemStorage(key, JSON.stringify(body), true);
 
     // If there is a callback, forward the response body
     if (callback) {
--- a/www_src/lib/platform.js
+++ b/www_src/lib/platform.js
@@ -1,0 +1,171 @@
+/**
+ * Device agnostic platform interface.
+ *
+ * @package core
+ * @author Andrew Sliwinski <a@mozillafoundation.org>
+ */
+
+var lru = require('lru-cache');
+
+/**
+ * Constructor
+ */
+function Platform () {
+  // Check to see if window.Platform exists. If it does, return it.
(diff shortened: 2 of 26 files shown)


www_src/lib/api.js
```diff
@@ -41,13 +43,11 @@ function api(options, callback) {
...
-  // Set cache key
+  // Caching
   var key = 'cache::' + options.method + '::' + options.uri;
-
-  // Use device cache if window.Android is available & options.useCache is true
-  if (window.Android && options.useCache === true && options.method === 'GET') {
+  if (options.useCache === true && options.method === 'GET') {
     console.log('Fetching from cache "' + key + '"');
-    var hit = window.Android.getMemStorage(key, true);
+    var hit = platform.getMemStorage(key, true);
```
thisandagain: Using the platform polyfill for API request caching

www_src/lib/render.jsx
```diff
@@ -14,8 +17,8 @@ var Base = React.createClass({
   onBackPressed: function() {
     if (this.state.onBackPressed) {
       return this.state.onBackPressed();
-    } else if (window.Android) {
-      window.Android.goBack();
+    } else {
+      platform.goBack();
```
thisandagain: Using the platform polyfill for detecting / handling the Android back button

www_src/pages/make/make.jsx
```diff
@@ -79,10 +80,9 @@ var Make = React.createClass({
       if (!body || !body.project) {
         return this.onEmpty();
       }
-      if (window.Android) {
-        window.Android.trackEvent('Make', 'Create a Project', 'New Project Started');
-        window.Android.setView('/users/' + user.id + '/projects/' + body.project.id);
-      }
+      
+      platform.trackEvent('Make', 'Create a Project', 'New Project Started');
```
thisandagain: Using the platform polyfill for the entire "Make" activity including GA tracking, view transitions, and user session reset.

www_src/lib/platform.js
```diff
@@ -0,0 +1,171 @@
...
+
+Platform.prototype.resetSharedPreferences = function() {
+  return this.sharedStorage.clear();
+};
+
+// -----------------------------------------------------------------------------
+// Session storage
+// -----------------------------------------------------------------------------
+
+Platform.prototype.getUserSession = function() {
```
Pomax: getSharedPreferences

www_src/lib/render.jsx
```diff
@@ -1,9 +1,12 @@
 var React = require('react');
+
 var Spindicator = require('../components/spindicator/spindicator.jsx');
 var ModalConfirm = require('../components/modal-confirm/modal-confirm.jsx');
 var ModalSwitch = require('../components/modal-switch/modal-switch.jsx');
 var Snackbar = require('../components/snackbar/snackbar.jsx');
 
+var platform = require('./platform');
```
gvn: Is there a reason that this is an explicit include, and elsewhere Platform is a global?

www_src/pages/make/make.jsx
```diff
@@ -2,6 +2,7 @@ var React = require('react');
 var api = require('../../lib/api');
 var router = require('../../lib/router');
 var dispatcher = require('../../lib/dispatcher');
+var platform = require('../../lib/platform');
```
gvn: Another instance of require vs. global...

www_src/lib/platform.js
```diff
@@ -0,0 +1,171 @@
...
+ * @author Andrew Sliwinski <a@mozillafoundation.org>
+ */
+
+var lru = require('lru-cache');
+
+/**
+ * Constructor
+ */
+function Platform () {
+  // Check to see if window.Platform exists. If it does, return it.
```
gvn: Is this basically ensuring Platform is a singleton?

www_src/lib/platform.js
thisandagain: `getUserSession` uses the shared preferences API to get and set the user session using `this.sessionKey`.

www_src/lib/render.jsx
thisandagain: This is one of the places where I migrated to using the polyfill. Everything should migrate to using the `platform` module eventually but in the interest of keeping this small and avoiding surface area for regressions only the items listed in the description were migrated.

www_src/lib/platform.js
thisandagain: As per the description, this checks to see if the platform (in our current case, Android) was bound an interface to `window.Platform`. If it has, the platform module returns it instead of the polyfill.

PR #2374, opened 2015-06-30
Provide config via webpack's DefinePlugin

toolness:
Currently, `www_src/config.js` is auto-generated in a way that's not very easy to work with, especially considering our imminent move to factoring out `www_src` into a separate npm module.

This PR simplifies our build process by using webpack's [`DefinePlugin`](http://webpack.github.io/docs/list-of-plugins.html#defineplugin) to inject the configuration into webpack bundles instead. One advantage is that `www_src/config.js` is now actually a real, maintainable file instead of an auto-generated one--although it can be loaded in two different contexts, which does complicate things a bit, but hopefully my comments have made things easier to understand.

In the future I'd like to somehow integrate gradle into this process so I can receive a hug from @thisandagain.


gvn:
I'm having problems with this. It seems like some code is malformed...

screen shot 2015-07-01 at 1 13 20 pm

gvn:
Let's re-visit this after we've migrated to using webmaker-core.

PR #2378, opened 2015-07-01
Fix crash/stall when entering color tinker mode.

### LINKED ISSUE #2297: entering tinker mode in photos stalls the app (blank screen)
On some phones, entering tinker mode shows a blank screen (for images).

toolness:
This fixes #2297 by adding fault tolerance to the color tinker mode in the case when the color is an empty string or somehow malformed. It does this by falling back to a default color, which is `rgba(0, 0, 0, 0)` by default, but can easily be changed via a prop.

Adding tests for this was a bit funky because we don't appear to have any test suites that have a DOM. I can help with that if you need, as we have such a thing on teach and it's not very hard to set up. For now I've added a few tests using React's [shallow rendering](https://facebook.github.io/react/docs/test-utils.html#shallow-rendering) and other [workarounds](http://simonsmith.io/unit-testing-react-components-without-a-dom/#testing-component-methods).

### DIFF
--- a/www_src/components/color-spectrum/color-spectrum.jsx
+++ b/www_src/components/color-spectrum/color-spectrum.jsx
@@ -89,14 +89,28 @@ var RGB = React.createClass({
 });
 
 var ColorSpectrum = React.createClass({
+  getDefaultProps: function() {
+    return {
+      defaultColor: 'rgba(0, 0, 0, 0)'
+    };
+  },
   updateColor: function (color) {
     // Normalize any updates to rgba
     this.props.onChange(color.rgbaString());
   },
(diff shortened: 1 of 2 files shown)


toolness:
I also noticed that the `ColorSpectrum` component doesn't seem to define [`propTypes`](https://facebook.github.io/react/docs/reusable-components.html#prop-validation), which I think can be very useful for self-documenting and debugging our code. I'm happy to add that if you like, although I'm not actually sure what props are required vs. optional.

thisandagain:
Nice. This is working well on device / all tests passing. R+

PR #2379, opened 2015-07-01
Fix onPause not returning the correct super type

ryanw-se:
/cc @thisandagain

Should fix unregistered errors that @Pomax was encountering yesterday.


thisandagain:
Great catch @ryanw-se

PR #2389, opened 2015-07-02
mkdirping for missing intermediate folders

gvn:


ashleygwilliams:
:+1: for linux, setting up a virtual box to test for windows

ashleygwilliams:
according to the `mkdirp` repo, it should support windows and only has one relevant open issue: https://github.com/substack/node-mkdirp/issues/56

as long as we keep a valid path there, we should be good. still gonna test, but i anticipate it should work.

PR #2391, opened 2015-07-06
Add JavaScript method to get the system locale

ryanw-se:
/cc @cadecairos @secretrobotron 

You can now use ```getSystemLocale()``` to return the default systems locale as a string format, ```<language>_<region>```. The end result should look like, ```en_US```.

### DIFF
--- a/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
+++ b/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
@@ -22,6 +22,8 @@ import org.mozilla.webmaker.activity.Element;
 import org.mozilla.webmaker.router.Router;
 import org.mozilla.webmaker.storage.MemStorage;
 
+import java.util.Locale;
+
 public class WebAppInterface {
 
     protected Context mContext;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
```diff
@@ -290,4 +292,14 @@ public void openExternalUrl(String url) {
...
     }
+
+    /**
+     * ----------------------------------------
+     * Get System Locale
+     * ----------------------------------------
+     */
+    @JavascriptInterface
+    public String getSystemLocale() {
+        return Locale.getDefault().toString();
```
Pomax: `Locale.getDefault().getLanguage();` rather than `toString()`? Better to be intentional about what we need than the string serialization of the entire object.

secretrobotron: Do they report the same thing? i.e. does `getLanguage` give us `xx_XX` ?

alicoding: Did quick research and it says that `Locale.getDefault().getLanguage()` will return `English` vs `en` so I don't think we want that?

secretrobotron: :+1:

ryanw-se:
Reopened after confirmation, that this code functions properly.

Pomax:
R+ from me and @alicoding after the checks for what Android yields as codes.

PR #2393, opened 2015-07-06
Change locale response underscores to hypens

ryanw-se:
Output: https://gist.github.com/ryanw-se/9d2a241a1e68ba4ec5bb

This change was put in place because the browser uses dashes and not underscores like the result currently being outputted.

### DIFF
--- a/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
+++ b/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
@@ -299,7 +299,7 @@ public class WebAppInterface {
      * ----------------------------------------
      */
     @JavascriptInterface
-    public String getSystemLocale() {
-        return Locale.getDefault().toString();
+    public String getSystemLanguage() {
+        return Locale.getDefault().toString().replace("_", "-");
     }
 }
--- a/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
+++ b/app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
@@ -300,6 +300,7 @@ public class WebAppInterface {
      */
     @JavascriptInterface
     public String getSystemLanguage() {
+        // Change underscores to dashes (The browser uses dashes instead)
         return Locale.getDefault().toString().replace("_", "-");
     }
 }


app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
```diff
@@ -300,6 +301,6 @@ public boolean isDebugBuild() {
      */
     @JavascriptInterface
     public String getSystemLocale() {
-        return Locale.getDefault().toString();
+        return Locale.getDefault().toString().replaceAll(String.valueOf(Pattern.compile("[_]")), "-");
```
Pomax: Since we're only replacing a delimited, I don't think we need the regexp, a straight up `...toString().replaceAll("_","-")` should do the trick

app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
```diff
@@ -300,6 +300,6 @@ public boolean isDebugBuild() {
      */
     @JavascriptInterface
     public String getSystemLocale() {
-        return Locale.getDefault().toString();
+        return Locale.getDefault().toString().replaceAll("_", "-");
```
Pomax: R+ if we add a oneline comment here to explain what the replace call achieves

ryanw-se: /cc @secretrobotron

alicoding: I also think this should now be change to `getSystemLanguage` since locale standard is underscore and language code is hyphen.

ryanw-se: @alicoding done!

ryanw-se:
This change was put in place because ```webmaker-browser``` uses dashes and not underscores like the result currently being outputted.

app/src/main/java/org/mozilla/webmaker/javascript/WebAppInterface.java
```diff
@@ -299,7 +299,7 @@ public boolean isDebugBuild() {
      * ----------------------------------------
      */
     @JavascriptInterface
-    public String getSystemLocale() {
-        return Locale.getDefault().toString();
+    public String getSystemLanguage() {
+        return Locale.getDefault().toString().replace("_", "-");
```
k88hudson: mind adding a comment here why this is necessary

ryanw-se: This change was put in place because the browser uses dashes and not underscores like the result currently being outputted.

ryanw-se:
R? @k88hudson

alicoding:
R+

toolness:
<a href="https://github.com/mozilla/webmaker-core/issues/277"></a>
This issue has been moved to https://github.com/mozilla/webmaker-core/issues/277.
Please don't comment on it here.

PR #2410, opened 2015-07-15
Make "npm run build" work on Windows.

toolness:
This fixes mozilla/webmaker-core#297 in a way slightly different from https://github.com/mozilla/webmaker-browser/pull/164: it adds a very limited drop-in replacement for `npm explore <name> -- <cmd>` at `npm_tasks/explore.js` that works on Windows. I did this because webmaker-android's `package.json` uses `npm explore` for multiple tasks. It also makes it easier to migrate back to a pure npm solution if/when https://github.com/npm/npm/issues/8932 is fixed.

If we really like this solution, we can potentially use it in webmaker-browser too, and use any of a variety of mechanisms to ensure we stay DRY.

### DIFF
--- a/npm_tasks/explore.js
+++ b/npm_tasks/explore.js
@@ -1,0 +1,34 @@
+// This is meant to provide a cross-platform way of doing the same thing
+// as "npm explore <name> -- <cmd>" (which doesn't work on Windows; see
+// https://github.com/npm/npm/issues/8932).
+//
+// Note that <cmd> is expected to be something extremely simple like
+// "npm install" or "git pull origin master". It shouldn't contain any
+// quoted arguments or other funky things.
+
+var fs = require('fs');
+var path = require('path');
+var execSync = require('child_process').execSync;
+
+var ROOT_DIR = path.resolve(__dirname, '..');
+
+function getModuleDir(moduleName) {
+  return fs.realpathSync(path.resolve(ROOT_DIR, 'node_modules', moduleName));
+}
+
+function explore(moduleName, cmd) {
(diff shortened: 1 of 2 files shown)


npm_tasks/explore.js
```diff
@@ -0,0 +1,34 @@
+// This is meant to provide a cross-platform way of doing the same thing
+// as "npm explore <name> -- <cmd>" (which doesn't work on Windows; see
+// https://github.com/npm/npm/issues/8932).
+//
+// Note that <cmd> is expected to be something extremely simple like
+// "npm install" or "git pull origin master". It shouldn't contain any
+// quoted arguments or other funky things.
```
toolness: This is an annoying limitation and we should probably get rid of it if we decide to turn this into an npm module or something. Um, assuming there isn't already an npm module that does the same thing. :eyes:

npm_tasks/explore.js
```diff
@@ -0,0 +1,34 @@
...
+var execSync = require('child_process').execSync;
+
+var ROOT_DIR = path.resolve(__dirname, '..');
+
+function getModuleDir(moduleName) {
+  return fs.realpathSync(path.resolve(ROOT_DIR, 'node_modules', moduleName));
+}
+
+function explore(moduleName, cmd) {
+  execSync(cmd, { cwd: getModuleDir(moduleName), stdio: 'inherit' });
```
Pomax: is there a way to turn this into `spawn` instead of `exec`, instead? The problem with exec is that all the output from that step is lost (there is not stdout/stderr piping)

PR #2415, opened 2015-07-20
google play string yml file for transifex

secretrobotron:
So we can upload to transifex, and simply copy & paste into play store when necessary.

Provides strings for https://github.com/mozilla/webmaker-core/issues/289

### DIFF
--- a/google-play.yml
+++ b/google-play.yml
@@ -1,0 +1,5 @@
+---
+en-US:
+  title: Webmaker Beta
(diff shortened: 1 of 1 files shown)


secretrobotron:
@alicoding r?

secretrobotron:
cc @thisandagain

secretrobotron:
After R+, I'll upload to transifex.

alicoding:
R+

PR #2416, opened 2015-07-27
Add NPM tasks to gradle build files

ryanw-se:
Adds the command ```npm run build``` to the gradle build files to run before the app is actually compiled. 

It also runs the command ```npm install``` if it has never been ran before, or if changes have been made to the ```package.json``` file or ```node_modules```.

Closes mozilla/webmaker-core#365, mozilla/webmaker-core#281, mozilla/webmaker-core#276

/cc @thisandagain

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.application'
+apply plugin: 'com.moowork.node'
 
 android {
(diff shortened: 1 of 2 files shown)


thisandagain:
Looks and works great. Tracking small follow-up with your changes in https://github.com/mozilla/webmaker-android/pull/2420

PR #2417, opened 2015-07-29
Extract config variables from strings.xml

alicoding:
Extracting these to its own config file, so we can use strings.xml for localization instead.


alicoding:
@k88hudson R?

ryanw-se:
This change would break Google Analytics support, I originally put them in strings.xml because those values needed to be referenced in the code itself. See, https://github.com/mozilla/webmaker-android/blob/develop/app/src/main/java/org/mozilla/webmaker/WebmakerApplication.java#L32

alicoding:
@k88hudson I also added all translation files into this patch as well :)

k88hudson:
@alicoding R+

PR #2427, opened 2015-08-04
Add isNetworkAvailable method [mozilla/webmaker-core#433]

k88hudson:
Although https://gist.github.com/thisandagain/0257dec9f36b376e36c2 would work it still relies on api.webmaker.org being up/available.

I saw this documented here: http://developer.android.com/training/monitoring-device-state/connectivity-monitoring.html#DetermineConnection, and it seems to work fine. What do you think @thisandagain?


ryanw-se:
Looks solid, :+1:

thisandagain:
Nice! This approach works too! :+1:

PR #2428, opened 2015-08-13
Master

satdav:
Here is the en-gb file for the app


thisandagain:
Will resolve in follow-up PR

PR #2432, opened 2015-08-19
changeView... methods instead of setView

Pomax:
Necessary for the refactor work in https://github.com/mozilla/webmaker-core/issues/451 and https://github.com/mozilla/webmaker-core/issues/452


ryanw-se:
This looks sharp.

PR #2434, opened 2015-08-25
[WIP] Merging MemStorage into our new cache system

### LINKED ISSUE #591: Replace "block editor" with a transition state
A lot of people find it difficult to understand the editing paradigm and visualize changes they were making whilst editing blocks.

We want to try transitioning the block to the top, fading out the rest of the app and adding editing UI on the same page.

Some things to consider:
* how does this work for large blocks? e.g. image
* how do we exit that state/initiate a "save"?

ryanw-se:
Closes mozilla/webmaker-core#591

This WIP includes a light refactor, changes were minimal which is why I didn't put them in their own PR, mostly conventional changes and of course a few to make things easier to read.


ryanw-se:
@Pomax R?

app/src/main/java/org/mozilla/webmaker/WebmakerApplication.java
```diff
@@ -6,7 +6,15 @@
 import com.google.android.gms.analytics.GoogleAnalytics;
 import com.google.android.gms.analytics.Tracker;
 
-import org.mozilla.webmaker.activity.*;
+import org.mozilla.webmaker.activity.Element;
```
Pomax: why not keep *?

app/src/main/java/org/mozilla/webmaker/activity/Element.java
```diff
@@ -5,13 +5,13 @@
...
 import android.util.Log;
 
-import java.io.File;
-import java.util.UUID;
-
 import org.mozilla.webmaker.R;
 import org.mozilla.webmaker.WebmakerActivity;
 import org.mozilla.webmaker.util.Image;
 
+import java.io.File;
```
Pomax: harmless, but we might as well revert this since the import order doesn't matter.

app/src/main/java/org/mozilla/webmaker/activity/Play.java
```diff
@@ -6,13 +6,14 @@
...
 import android.view.MenuItem;
+
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.mozilla.webmaker.R;
 import org.mozilla.webmaker.WebmakerActivity;
-import org.mozilla.webmaker.web.javascript.WebAppInterface;
 import org.mozilla.webmaker.router.Router;
 import org.mozilla.webmaker.util.Share;
+import org.mozilla.webmaker.web.javascript.WebAppInterface;
```
Pomax: again probably better to just revert these changes, since they aren't relevant to the patch itself, and don't actually change anything in how things are compiled.

app/src/main/java/org/mozilla/webmaker/web/javascript/WebmakerAPI.java
```diff
@@ -1,22 +1,21 @@
...
 public class WebmakerAPI {
 
-    // Static behaviour
+    // used for triggering JS on the main UI thread
+    protected WebmakerWebView currentView = null;
+    protected BaseActivity currentActivity = null;
 
+    protected Queue queue;
+
+    // Static behaviour
```
Pomax: this was organized as "statics first, then instance", so if we're going to mix the two, the comments about which block starts where is no longer really relevant, I'd recommend keeping the static vs instance functionality grouped rather than mixed, since the two blocks aren't supposed to casually interact.

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -129,7 +128,8 @@ public String getMemStorage(String key) {
     @JavascriptInterface
     public String getMemStorage(String key, final boolean global) {
         if (!global) key = key.concat(mPrefKey);
-        return MemStorage.sharedStorage().get(key);
+        //return MemStorage.sharedStorage().get(key);
+        return WebmakerAPI.instance.getPayloads(key);
```
Pomax: we bind the API instance as `api` on https://github.com/ryanw-se/webmaker-android/blob/migrate-cache/app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java#L56 so we should be able to just do `return api.getPaylods(key)` rather than having to tap into the static `instance` variable on the `WebmakeAPI` class

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -140,7 +140,8 @@ public void setMemStorage(String key, final String value) {
     @JavascriptInterface
     public void setMemStorage(String key, final String value, final boolean global) {
         if (!global) key = key.concat(mPrefKey);
-        MemStorage.sharedStorage().put(key, value);
+        //MemStorage.sharedStorage().put(key, value);
+        WebmakerAPI.instance.queue(key, value);
```
Pomax: `api.queue`

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -229,7 +230,8 @@ public void setView(final String url, final String routeData) {
         if (activity == null) return;
 
         if (routeData != null) {
-            MemStorage.sharedStorage().put(ROUTE_KEY, routeData);
+            //MemStorage.sharedStorage().put(ROUTE_KEY, routeData);
+            WebmakerAPI.instance.queue(ROUTE_KEY, routeData);
```
Pomax: `api.queue`

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -254,12 +256,14 @@ public String getRouteParams() {
 
     @JavascriptInterface
     public String getRouteData() {
-        return MemStorage.sharedStorage().get(ROUTE_KEY);
+        //return MemStorage.sharedStorage().get(ROUTE_KEY);
+        return WebmakerAPI.instance.getPayloads(ROUTE_KEY);
```
Pomax: `api.getPayloads`

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -254,12 +256,14 @@ public String getRouteParams() {
...
     }
 
     @JavascriptInterface
     public void clearRouteData() {
-        MemStorage.sharedStorage().put(ROUTE_KEY, "");
+        //MemStorage.sharedStorage().put(ROUTE_KEY, "");
+        WebmakerAPI.instance.queue(ROUTE_KEY, "");
```
Pomax: `api.queue`

PR #2435, opened 2015-08-26
[Squashed] Begin work on merging MemStorage into our new cache system

ryanw-se:
This is a compressed version of PR #2434. @Pomax R?


Pomax:
You don't need to file a new PR for squashed commits: if you do a rebase, you can then simply update the branch you already had with a `git push yourrepo yourbranch -f` and it will update the branch properly, with github automatically reflecting that to your open PR.

ryanw-se:
@Pomax Neato!

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -129,7 +128,7 @@ public String getMemStorage(String key) {
     @JavascriptInterface
     public String getMemStorage(String key, final boolean global) {
         if (!global) key = key.concat(mPrefKey);
-        return MemStorage.sharedStorage().get(key);
+        return api.getPayloads(key);
```
Pomax: since the queue system does not automatically remove queued items, and returns the list of items queued, rather than a single item, we will need to make sure to extract the element, and clear the queue:
```
String cachedItem = api.getPayload(key, 0);
api.clearPayloads(key);
return cachedItem;
```

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -140,7 +139,7 @@ public void setMemStorage(String key, final String value) {
     @JavascriptInterface
     public void setMemStorage(String key, final String value, final boolean global) {
         if (!global) key = key.concat(mPrefKey);
-        MemStorage.sharedStorage().put(key, value);
+        api.queue(key, value);
```
Pomax: the old MemStorage calls overwrite already cached data, which the new system doesn't do: it just queues "a next data payload", so we want to make sure to clear a queue before putting the data in, so that it's the only element:
```
api.clearPayloads(key);
api.queue(key, value);
```

xmatthewx:
Closes https://github.com/mozilla/webmaker-core/issues/591

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -229,7 +228,7 @@ public void setView(final String url, final String routeData) {
         if (activity == null) return;
 
         if (routeData != null) {
-            MemStorage.sharedStorage().put(ROUTE_KEY, routeData);
+            api.queue(ROUTE_KEY, routeData);
```
Pomax: if we don't add a `setPayload` method, this will need to make sure we clear first, too. Otherwise, this should become `setPayload`.

app/src/main/java/org/mozilla/webmaker/web/javascript/WebAppInterface.java
```diff
@@ -254,12 +253,12 @@ public String getRouteParams() {
...
     @JavascriptInterface
     public String getRouteData() {
-        return MemStorage.sharedStorage().get(ROUTE_KEY);
+        return api.getPayloads(ROUTE_KEY);
     }
 
     @JavascriptInterface
     public void clearRouteData() {
-        MemStorage.sharedStorage().put(ROUTE_KEY, "");
+        api.queue(ROUTE_KEY, "");
```
Pomax: there is a clear method already, so this can be turned into `api.clearPayloads(ROUTE_KEY)`

k88hudson:
@ryanw-se @Pomax let's land this in the next release (1.4)

ryanw-se:
@k88hudson When is the approximate 1.4 landing date?

ryanw-se:
I'm going to pick this up at a later date, need to do some other work before I get to this, will make life much easier when I come back to this as well.

PR #2440, opened 2015-09-16
Upgrade the router, take advantage of caching functionality

ryanw-se:
The goal was simple, I wanted to create a routing solution that contained the features we need, without the stuff we don't need, while still being just as functional as it was before. I also focused on making everything extremely simple and easy to understand and everything is heavily documented in each file.

Here are a couple of things you will notice, when comparing this solution to the old router.

- Router methods were renamed to better describe what they do, route routes a url, etc.
- Routes are defined as a new route object. Predefined routes are partial route objects, called routes are fully defined route objects with parameters included. The data in predefined routes is copied to called routes.
- Caching now functions properly, previously we didn't clean the URL when checking if the URL was cached, so we would put stuff in the cache but never use it. :octocat: 
- A much more simplistic design, a lot of unnecessary logic was removed, with the goal of making everything as simple as possible. This is what improves the performance in comparison to the old solution. 

I refactored a version specifically for webmaker-android, however you can view the base project here, https://github.com/ryanw-se/SimpleRouter.


k88hudson:
This is awesome! :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: 

Looks good to me. I say we get this out to the beta channel :+1:

PR #2444, opened 2015-09-21
strings.xml

bolaram:
Some spellings updated :)


alicoding:
Hi @bolaram Thank you so much for the PR. I'm going to merge this PR, but I would like to ask you to also update the string on Transifex here https://www.transifex.com/webmaker/webmaker/webmaker-stringsxml/ so that it won't override what's in here if someone is going to change them on Transifex.

bolaram:
@alicoding thanks. i updated string on  https://www.transifex.com/webmaker/webmaker/webmaker-stringsxml/ .. i just added string suggestion there :) will you check that ?

PR #2445, opened 2015-09-21
switching to node 4

gvn:


ryanw-se:
@alanmoo Can this be merged? It appears we're already using Node 4 in webmaker-core (https://github.com/mozilla/webmaker-core/blob/develop/package.json#L35).

alanmoo:
Taken care of by #2452

PR #840, opened 2015-01-29
Fix #803 v-bind-i18n-html

pippinlee:
Update `v-bind-html` directive and documentation to be `v-bind-i18n-html`. The goal is to help make it easier to understanding that it passes the contents through the `i18n` filter.

PR #2449, opened 2015-11-16
adding Web Link view

gvn:
Enables https://github.com/mozilla/webmaker-core/pull/802

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -124,5 +124,11 @@
             android:label="@string/title_activity_style_guide"
             android:screenOrientation="portrait" >
         </activity>
+
+        <activity
+            android:name=".activity.WebLink"
+            android:label="@string/title_activity_web_link"
+            android:screenOrientation="portrait" >
+        </activity>
     </application>
 </manifest>
--- a/app/src/main/java/org/mozilla/webmaker/WebmakerApplication.java
+++ b/app/src/main/java/org/mozilla/webmaker/WebmakerApplication.java
@@ -44,7 +44,9 @@ public class WebmakerApplication extends Application {
 
         SimpleRouter.getSharedRouter().route("/style-guide", StyleGuide.class);
 
-	SimpleRouter.getSharedRouter().route("/users/:user/projects", UserProjects.class);
+        SimpleRouter.getSharedRouter().route("/web-link", WebLink.class);
+
+        SimpleRouter.getSharedRouter().route("/users/:user/projects", UserProjects.class);
         SimpleRouter.getSharedRouter().route("/users/:user/projects/:project", Project.class);
         SimpleRouter.getSharedRouter().route("/users/:user/projects/:project/settings", ProjectSettings.class);
         SimpleRouter.getSharedRouter().route("/users/:user/projects/:project/play", Play.class);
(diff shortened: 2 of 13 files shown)


alanmoo:
These could apply to https://github.com/mozilla/webmaker-core/pull/789 as well

* When in the view to add a link to web, I can't back out of it unless I have content
* Should the validation message update as the user types?
* I'm assuming it's by design that the initial "Set Destination" action on buttons is only for within the project? Seemed a little odd at first but I understand why we might do that- web links are an advanced feature, right?

There's a bug that I'm having trouble replicating in which the first page no longer becomes tappable in the project overview. It seems to get triggered with some combination of creating a link, tapping the initial "set destination" and then backing out if it, but I can't pinpoint it.  The page becomes greyed out in the "set destination view" and remains like that even in the "add page" view, like the rest of the "add page" placeholders around it. It stays this way even after adding more pages, and the only way to change it is to delete all other pages and then you can go in to edit mode, which defaults to that page.

gvn:
> When in the view to add a link to web, I can't back out of it unless I have content

Fixed.

> Should the validation message update as the user types?

Yep, fixed.

>I'm assuming it's by design that the initial "Set Destination" action on buttons is only for within the project? Seemed a little odd at first but I understand why we might do that- web links are an advanced feature, right?

Yeah, I think we settled on leaving that as-is for this round of implementation and seeing how it feels.

gvn:
Fixups are in: https://github.com/mozilla/webmaker-core/pull/789

gvn:
As to that last bug...let me know if you get solid repro steps. Maybe file a follow-up if not...

alanmoo:
Sounds good. Figured the greying of the page might ring a bell, but I'll play around with it some more when I review this.

alanmoo:
After setting a link to an external URL, the button still has the "Set Destination" button below it in the edit page view. Not sure if this is desired or not.

After following a button to an external URL from the first page of a project, returning to Webmaker results in an inability to use the arrow button to change pages within the project unless you zoom out and in again.

After following a button to an external URL from another page in the project, switching back to Webmaker and results in the project returning to its first page.

xmatthewx:
> switching back to Webmaker and results in the project returning to its first page. 

This refresh happens other times you leave the app. Always a little annoying, worse in this scenario. @gvn should we file this a separate issue?

> the button still has the "Set Destination" button below it 

We might want to remove the "set destination" helper button. Since "set destination" helper doesn't offer URL as an option, it risks more harm than good. "Follow link" helper should remain. What do you think?

gvn:
> switching back to Webmaker and results in the project returning to its first page. 

>> This refresh happens other times you leave the app. Always a little annoying, worse in this scenario. @gvn should we file this a separate issue?

Yeah, if we could do that separately it would help move things forward.

> the button still has the "Set Destination" button below it 

>> We might want to remove the "set destination" helper button. Since "set destination" helper doesn't offer URL as an option, it risks more harm than good. "Follow link" helper should remain. What do you think?

Sounds good to me. I'll make that change...

xmatthewx:
filed #840 on auto-refresh projects

alanmoo:
> After following a button to an external URL from the first page of a project, returning to Webmaker results in an inability to use the arrow button to change pages within the project unless you zoom out and in again.

I'm still getting this @gvn. Is this part of #840 as well?

xmatthewx:
Unrelated to @alanmoo's question ... but in case you didn't notice, we'd like to add an icon to links that go to URLs, so people viewing a project can distinguish. https://github.com/mozilla/webmaker-core/issues/797

PR #2452, opened 2015-12-15
Upgrade app dependencies, Switch to Node 4

ryanw-se:
The hype train is now mobile! :steam_locomotive:


alanmoo:
@ryanw-se, think it would be best to throw in the changes to the readme and `.travis.yml` from #2445 so we can just close that one without merging?

package.json
```diff
@@ -11,22 +21,16 @@
...
     "webmaker-core": "https://github.com/mozilla/webmaker-core.git#6c1b8afbe0c1c02af750c667f7bafcf6215c085a"
   },
-  "homepage": "https://github.com/mozilla/webmaker-android",
   "devDependencies": {
     "mkdirp": "^0.5.1",
     "ncp": "^2.0.0",
     "rimraf": "^2.4.1"
+  },
+  "engines": {
+    "node": "^4.1.0",
```
ryanw-se: This is here to prevent people from attempting to run NPM commands on webmaker-android, with an outdated version of NPM or Node outside of the Gradle build environment. If you trigger a Gradle build or use any commands with the Gradle prefix (I.e. gradle npmInstall), it will use the Node version Gradle downloaded.

app/build.gradle
```diff
@@ -76,14 +76,14 @@ repositories {
...
-    compile 'org.xwalk:xwalk_core_library:13.42.319.12'
+    compile 'com.android.support:support-v13:23.1.1'
+    compile 'com.android.support:appcompat-v7:23.1.1'
+    compile 'com.google.android.gms:play-services-analytics:8.3.0'
+    compile 'org.xwalk:xwalk_core_library:15.44.384.12'
 }
 
 node {
-    version = '0.12.2'
+    version = '4.2.3'
```
ryanw-se: This just specifies which version of Node we want the mooworks plugin to download and use to compile the app.

gradle.properties
```diff
@@ -15,4 +15,6 @@
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
\ No newline at end of file
+# org.gradle.parallel=true
+
+android.useDeprecatedNdk=true
```
ryanw-se: With the upgraded android build tools, you have to add the "android.useDeprecatedNdk=true" flag in order to continue using the NDK tools with newer versions of build tools. Ultimately, they've deprecated this and are trying out a new experimental build system that I don't believe we are quite ready to switch over to yet.

app/build.gradle
```diff
@@ -2,13 +2,13 @@ apply plugin: 'com.android.application'
...
 android {
-    compileSdkVersion 22
+    compileSdkVersion 23
     buildToolsVersion "21.1.2"
 
     defaultConfig {
         applicationId "org.mozilla.webmaker"
         minSdkVersion 14
-        targetSdkVersion 22
+        targetSdkVersion 23
```
ryanw-se: Its usually a good idea to always target the latest version of Android, whilst still providing support down to the lowest point possible (SDK 14), I had to update this because the dependency updates below were complaining rather loudly about it.

alanmoo:
Had to download new versions of the SDK on mediocre wi-fi, but it looks good! 🎉

PR #2460, opened 2016-01-11
Remove empty XML tags from string resources & Implement permanent solution for TagList view header

ryanw-se:
Also brings the AndroidManifest up to snuff with standard XML convention (and fixes weird formatting).

Related: mozilla/webmaker-android#2459, mozilla/webmaker-core#975

cc: @gvn, @xmatthewx


gvn:
Looks good!

